PR #58, opened 2025-10-15
fix: Using sqlite, panic occurred during deployment via docker compose

RookieZoe:
fix issue: https://github.com/looplj/axonhub/issues/57

PR #59, opened 2025-10-15
fix: After changing the language preference, it reverted after refres…

RookieZoe:
Root cause

In the SQLite database environment, the reason why preferLanguage does not take effect immediately after being updated is:

1. Cache not invalidated: The UpdateMe resolver (internal/server/gql/me.resolvers.go:20-52) does not clear the UserService cache after directly updating the database.

2. Cache hit: Subsequent requests still return the old preferLanguage value when reading from the cache through GetUserByID (internal/server/biz/user.go:139-172).

3. Restart takes effect: Only after restarting the service and clearing the cache can the new value be read from the database.

### DIFF
--- a/internal/server/biz/user.go
+++ b/internal/server/biz/user.go
@@ -112,7 +112,7 @@ func (s *UserService) UpdateUser(ctx context.Context, id int, input ent.UpdateUs
 	}
 
 	// Invalidate cache
-	s.invalidateUserCache(ctx, id)
+	s.InvalidateUserCache(ctx, id)
 
 	return user, nil
 }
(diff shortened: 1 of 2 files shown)


looplj:
Thanks for your PR, it should be duplicated with the last PR #58.
I have merged it.

RookieZoe:
Sorry about that, Because I want one PR one Commit. I'll close this, thanks.

looplj:
It's the best practice to keep a PR single-responsibility, and it is fine if the changes are minimal and a one-line commit message can clearly describe them.

Thanks for reporting and fixing the bug.

PR #168, opened 2025-11-26
feat: support override request headers

looplj:


looplj:
@gemini-code-assist  review

PR #209, opened 2025-12-02
fix: query should be passed through to the upstream service

looplj:


looplj:
@gemini-code-assist review

PR #221, opened 2025-12-05
feat: add extra body for gemini openai compatibility API

looplj:


looplj:
@gemini-code-assist review

PR #288, opened 2025-12-14
opt: bulk operation ui & apply override template

looplj:


looplj:
@gemini-code-assist review

PR #308, opened 2025-12-16
fix: anthropic thinking and responses convert

looplj:


looplj:
@gemini-code-assist  review

PR #331, opened 2025-12-21
opt: use global cached regexp package

looplj:


looplj:
@gemini-code-assist review

PR #541, opened 2026-01-10
fix: correct Gemini token count conversion to prevent negative candidatesTokenCount

mcowger:
## Problem

When AxonHub received a response from the chat completions API and transformed it to Gemini format, it incorrectly calculated `candidatesTokenCount` by subtracting `thoughtsTokenCount` from `completionTokens`. This caused negative values when reasoning tokens exceeded completion tokens.

## Reproduction Case

**Input (LLM/OpenAI format):**
```json
{
  "completion_tokens": 64,
  "total_tokens": 1761,
  "prompt_tokens": 1697,
  "completion_tokens_details": {
    "reasoning_tokens": 253
  }
}
```

**Output (Gemini format) - BEFORE FIX:**
```json
{
  "promptTokenCount": 1697,
  "candidatesTokenCount": -189,     ← WRONG: 64 - 253 = -189
  "totalTokenCount": 1761,
  "thoughtsTokenCount": 253
}
```

## Root Cause

The bug was based on a misunderstanding of token semantics in both formats:

- **LLM/OpenAI format**: `completion_tokens` does NOT include `reasoning_tokens` (they are separate, non-overlapping values)
- **Gemini format**: `candidatesTokenCount` does NOT include `thoughtsTokenCount` (they are also separate, non-overlapping values)

The old code incorrectly did:
```go
usage.CandidatesTokenCount = chatUsage.CompletionTokens - usage.ThoughtsTokenCount
```

## Solution

Changed the conversion to map `CompletionTokens` directly to `CandidatesTokenCount`:

```go
usage.CandidatesTokenCount = chatUsage.CompletionTokens
```

**Output (Gemini format) - AFTER FIX:**
```json
{
  "promptTokenCount": 1697,
  "candidatesTokenCount": 64,      ← CORRECT: direct mapping
  "totalTokenCount": 1761,
  "thoughtsTokenCount": 253
}
```

## Changes

- `llm/transformer/gemini/convert.go`: Fixed token calculation with detailed comments explaining the correct semantics
- `llm/transformer/gemini/inbound_convert_test.go`: Updated existing test expectations and added regression test case using the actual bug reproduction data

## Test Plan

All existing tests pass, including:
- Updated test case with corrected expectations
- New regression test case that specifically tests the scenario where reasoning tokens exceed completion tokens (preventing future regressions)

```bash
go test ./llm/transformer/gemini/...
```

### DIFF
--- a/frontend/src/features/requests/components/request-detail-page.tsx
+++ b/frontend/src/features/requests/components/request-detail-page.tsx
@@ -220,8 +220,10 @@ export default function RequestDetailPage() {
               const writeCachedTokens = usage.promptWriteCachedTokens || 0;
               const hasReadCache = cachedTokens > 0;
               const hasWriteCache = writeCachedTokens > 0;
-              const cacheHitRate = hasReadCache ? ((cachedTokens / promptTokens) * 100).toFixed(1) : '0.0';
-              const writeCacheRate = hasWriteCache ? ((writeCachedTokens / promptTokens) * 100).toFixed(1) : '0.0';
+              // Cache hit rate = cached tokens / (prompt tokens + cached tokens) * 100
+              // This represents the percentage of prompt tokens that were served from cache
+              const cacheHitRate = hasReadCache ? ((cachedTokens / (promptTokens + cachedTokens)) * 100).toFixed(1) : '0.0';
+              const writeCacheRate = hasWriteCache ? ((writeCachedTokens / (promptTokens + writeCachedTokens)) * 100).toFixed(1) : '0.0';
 
               return (
                 <Card className='border-0 shadow-sm'>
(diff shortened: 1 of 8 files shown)


mcowger:
> ## Code Review
> This pull request correctly fixes a bug that caused negative `candidatesTokenCount` during LLM-to-Gemini format conversion. The logic change is sound, and the addition of detailed comments and a regression test is commendable.
> 
> However, this fix highlights a critical inconsistency in the reverse conversion. The function `convertToLLMUsage` (in `llm/transformer/gemini/convert.go`, starting at line 310) still operates on the old, incorrect assumption that `llm.Usage.CompletionTokens` should be inclusive of reasoning tokens. Specifically, line 333 (`usage.CompletionTokens += geminiUsage.ThoughtsTokenCount`) contradicts the token semantics clarified in this PR, which state that `completion_tokens` and `reasoning_tokens` are separate, non-overlapping values in the LLM/OpenAI format.
> 
> This will result in incorrect token counts when converting from Gemini to the LLM format. To ensure consistency, it is highly recommended to remove the addition on line 333 as part of this PR.

Good catch gemini!   Fixed in updated commit.

looplj (approved):

PR #585, opened 2026-01-14
refactor: fix retry pipeline & retry on channels instead of models

looplj:


looplj:
@gemini-code-assist  review

PR #628, opened 2026-01-20
feat: add ordered load balancing strategy

### LINKED ISSUE #539: Adjustments to Adaptive Load Balancing
Before I go write a PR to do this - are you open to the idea that admins could adjust the weights used by the adapter load balancer? 

For example, the +1000 boost for same channel isn't what I want.  In fact, for my use cases, I'd like it to be very small. 

But I also don't want to go purely weight based (the other option) because the other components of the ALB (error rate etc) are desirable.  

Are you open to a PR that makes these configurable?

mcowger:
## Summary

Adds a new **ordered** load balancing strategy that provides strict, predictable channel prioritization based on definition order. This strategy always prefers the first channel and only uses subsequent channels when the previous ones fail.

This resolves #539 by providing an alternative to the adaptive load balancer for users who need deterministic channel selection without the complex weighting behavior of the adaptive strategy.

## Changes

- **New OrderedStrategy**: Implements simple ordering logic using `baseScore + orderingWeight`
  - Higher weight channels are always preferred
  - No dynamic adjustments based on success rates or trace affinity
  - Strict deterministic behavior for predictable routing
- **Comprehensive Tests**: 268 lines of unit tests covering edge cases, ordering guarantees, and consistency
- **Frontend Integration**: Added UI option for ordered strategy in retry settings
- **Internationalization**: Added English and Chinese translations
- **Orchestrator Integration**: Wired ordered load balancer with error-aware strategy for production use

## Behavior

The ordered strategy uses a straightforward scoring mechanism:
- Score = 1,000,000 (base) + orderingWeight
- Channels with higher weight get higher scores and are used first
- Unlike adaptive LB, there's no +1000 boost for same-channel affinity
- Unlike weighted LB, error-aware strategy is still applied to skip failing channels

## Use Case

This strategy is ideal for scenarios where you want:
- Channel A to always be tried first
- Channel B only as a fallback when A fails
- Predictable, deterministic routing without adaptive behavior
- Simple ordering without complex weighting calculations

## Test Plan

- [x] Unit tests pass for OrderedStrategy (268 LOC coverage)
- [x] Strategy correctly orders channels by weight (higher weight = higher priority)
- [x] Integration with orchestrator works correctly
- [x] Frontend UI displays ordered option in dropdown
- [x] Translations present in both English and Chinese

Resolves #539

## Summary by CodeRabbit

* **New Features**
  * Added "Ordered" load balancer strategy for retry configuration, enabling deterministic channel prioritization and an option in the retry dropdown.
  * Localized UI labels and documentation for the new strategy in English and Chinese.
  * Integrated the ordered strategy into orchestrator runtime selection.

* **Tests**
  * Added comprehensive unit tests validating ordered strategy scoring, debug output, and ordering guarantees.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


looplj:
It seems to duplicate with the weight strategy。

mcowger:
> It seems to duplicate with the weight strategy。

It doesn't.  The weight strategy it not deterministic.  If I set modelA = 90, model B = 10, out of 100 requests, approx 90 will go to A, and 10 will go to B.  

In the 'ordered' strategy, 100 requests would go to A, unless A is on a cooldown or disabled.

looplj:
It is a naming mistaken for the weight.

You can review the code,  the weight strategy only consider the weight instead of weighted round robin.

Maybe you can change the name to ordered, and keep the compatibility。

looplj:
It is a naming mistake for the weight.

You can review the code,  the weight strategy only consider the weight instead of weighted round robin.

Maybe we can change the name to ordered, and keep the compatibility。

mcowger:
> It is a naming mistaken for the weight.
> 
> You can review the code, the weight strategy only consider the weight instead of weighted round robin.
> 
> Maybe we can change the name to ordered, and keep the compatibility。

I dont think thats the case:

`internal/server/orchestrator/lb_strategy_rr.go`
```go
 weightFactor := float64(weight) / 100.0
  if weightFactor <= 0 {
      weightFactor = 1.0
  }
  normalizedCount := effectiveCount / weightFactor
  ```
  
  It doesn't enforce an ordering, the weight simply influences the selection likelihood.

File: internal/server/orchestrator/orchestrator.go [hunk 1]
```diff
@@ -38,6 +38,7 @@ func NewChatCompletionOrchestrator(
 
 	adaptiveLoadBalancer := NewLoadBalancer(systemService, channelService, strategies...)
 	weightedLoadBalancer := NewLoadBalancer(systemService, channelService, NewWeightStrategy())
```
looplj, line 40: The weightedLoadBalancer uses NewWeightStrategy instead of NewWeightRoundRobinStrategy

mcowger: OK, I see it now.  I'll close this one and submit a different PR to rename it to be more accurate.

PR #640, opened 2026-01-22
Update: Claude Code OAuth and Enhanced Cloaking

mcowger:
## Summary

Implements full Claude Code OAuth support and fixes tool prefix handling in responses.

### OAuth Implementation (commit 8668f3f8)
- **Frontend**: OAuth flow UI in channel configuration dialog
- **Backend**: OAuth endpoints for authorization flow
  - `/api/v1/claudecode/oauth/start` - Initiate OAuth with PKCE
  - `/api/v1/claudecode/oauth/callback` - Handle callback and exchange code for tokens
  - `/api/v1/claudecode/oauth/poll` - Poll for authorization status
- **Token Management**: Automatic token refresh using refresh tokens
- **Transformer**: OAuth token provider integration for Claude Code channels

### Tool Prefix Fix (commit d80b1d3b)
- Strip `proxy_` prefix from tool names in streaming and non-streaming responses
- Apply prefix only for OAuth tokens from non-CLI clients
- Track prefix application in request metadata for response processing
- Use correct authentication (x-api-key vs Bearer) based on endpoint

### Implementation Details
- `internal/server/api/claudecode.go`: OAuth flow handlers with PKCE
- `llm/transformer/anthropic/claudecode/token.go`: Token provider with refresh
- `llm/transformer/anthropic/claudecode/outbound.go`: Request transformation
- `llm/transformer/anthropic/claudecode/utils.go`: Utility functions
- `llm/transformer/anthropic/inbound_stream.go`: Streaming response prefix stripping
- `llm/transformer/anthropic/inbound_convert.go`: Non-streaming response prefix stripping

### Tests
- Fixed compilation errors (FunctionDefinition → Function, ToolChoice struct)
- Updated test expectations for authentication and system prompts
- All tests pass with 0 linting issues

## Fixes

- OAuth tokens are now properly obtained and refreshed for Claude Code channels
- Tool calls no longer fail with "No such tool available: proxy_bash" errors
- Correct authentication headers based on endpoint type

## Summary by CodeRabbit

* **New Features**
  * Claude Code OAuth PKCE sign-in with start/open/exchange flow and automatic credential import; Claude Code API format blocks added alongside Codex.

* **Refactor**
  * Unified, reusable OAuth flow powering Codex and Claude Code for consistent UX and state handling.

* **Localization**
  * English and Simplified Chinese strings added for Claude Code OAuth UI.

* **Bug Fixes / UX**
  * Improved Enter key behavior in dialog forms.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


frontend/src/features/channels/components/channels-action-dialog.tsx
```diff
@@ -111,6 +112,12 @@ export function ChannelsActionDialog({ currentRow, duplicateFromRow, open, onOpe
   const [isCodexStarting, setIsCodexStarting] = useState(false);
   const [isCodexExchanging, setIsCodexExchanging] = useState(false);
 
+  const [claudecodeSessionId, setClaudecodeSessionId] = useState<string | null>(null);
+  const [claudecodeAuthUrl, setClaudecodeAuthUrl] = useState<string | null>(null);
+  const [claudecodeCallbackUrl, setClaudecodeCallbackUrl] = useState('');
+  const [isClaudecodeStarting, setIsClaudecodeStarting] = useState(false);
+  const [isClaudecodeExchanging, setIsClaudecodeExchanging] = useState(false);
```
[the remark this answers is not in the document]
mcowger: This is fair.  Lemme do that.

llm/transformer/anthropic/claudecode/token.go
```diff
@@ -0,0 +1,585 @@
...
+		oauthUrls:   DefaultTokenURLs,
+		userAgent:   UserAgent,
+		creds:       params.Credentials,
+		onRefreshed: params.OnRefreshed,
+	}
+}
+
+// Exchange performs OAuth2 authorization_code exchange and returns credentials.
+// Claude uses JSON format instead of form-encoded data.
+func (p *ClaudeTokenProvider) Exchange(ctx context.Context, params oauth.ExchangeParams) (*oauth.OAuthCredentials, error) {
```
looplj: Can Claude utilize the shared OAuth TokenProvider? We recommend reusing the TokenProvider whenever possible, unless there are compatibility issues.

llm/transformer/anthropic/inbound_stream.go
```diff
@@ -355,12 +356,16 @@ func (s *anthropicInboundStream) Next() bool {
 						s.contentIndex += 1
 					}
 
+					// Strip proxy_ prefix from tool name if present
+					toolName := deltaToolCall.Function.Name
```
looplj: Will this cause the client to be unable to find the tool?

llm/transformer/anthropic/claudecode/token.go
mcowger: Yeah Gemini brought up above.  I think I can refactor to do that.

llm/transformer/anthropic/inbound_stream.go
mcowger: No, because it's only stripped when it was added previously.   

I've tested this most of the day with both opencode and Claude code and it hasn't been an issue.  This is also the same technique CPA uses.

mcowger:
## Addressed all PR review comments

**Human Reviewer (looplj):**
1. Token Provider refactoring → Refactored to use shared oauth.TokenProvider with ExchangeStrategy pattern for JSON token exchange
2. Tool prefix stripping → Already tested extensively with both opencode and Claude Code, using same approach as CPA

**Bot Comments:**
- Gemini: Token duplication → Fixed by refactoring to shared OAuth provider
- Gemini: Frontend OAuth duplication → Already using useOAuthFlow hook 
- Gemini: Redundant validation → Removed duplicate validation (binding tags handle it)
- CodeRabbit: Model identifiers → Updated to full date-suffixed IDs (claude-haiku-4-5-20251001, etc)
- CodeRabbit: URL parsing failure → Added fallback to Bearer auth when parsing fails
- CodeRabbit: Empty baseBetas bug → Fixed using strings.Join to prevent leading comma
- CodeRabbit: Backwards compatibility → Removed deprecated aliases per requirements
- TypeScript warnings → Fixed null handling and replaced deprecated onKeyPress with onKeyDown

All changes committed and pushed.

llm/transformer/anthropic/claudecode/token.go
mcowger: Refactored to use shared oauth.TokenProvider with ExchangeStrategy pattern for JSON token exchange.

llm/transformer/anthropic/inbound_stream.go
mcowger: Thoroughly tested with both opencode and Claude Code, using same approach as CPA.

llm/transformer/anthropic/claudecode/outbound.go
```diff
@@ -0,0 +1,328 @@
...
+	} else {
+		isAnthropicBase := strings.EqualFold(parsedURL.Scheme, "https") &&
+			strings.EqualFold(parsedURL.Host, "api.anthropic.com")
+
+		// For api.anthropic.com with API key: use x-api-key header
+		// For custom endpoints or OAuth tokens: use Bearer token
+		if isAnthropicBase && !isClaudeOAuthToken(apiKey) {
+			httpReq.Headers.Del("Authorization")
+			httpReq.Headers.Set("X-Api-Key", apiKey)
+			httpReq.Auth = &httpclient.AuthConfig{
```
looplj: Does the Claude code always require AuthTypeBearer?  

For API key authentication, the Anthropic channel type should be used.

We don't need to set headers here, as the HttpClient will automatically set them based on httpReq.Auth.

llm/transformer/anthropic/claudecode/utils.go
```diff
@@ -0,0 +1,296 @@
...
+	return "user_" + hexPart + "_account__session_" + uuidPart
+}
+
+// isValidUserID checks if a user ID matches Claude Code format.
+func isValidUserID(userID string) bool {
+	return userIDPattern.MatchString(userID)
+}
+
+// injectFakeUserID generates and injects a fake user ID into the request metadata.
+func injectFakeUserID(payload []byte) []byte {
```
looplj: Does the fake id will affect the input cache ?

llm/transformer/anthropic/inbound_stream.go
```diff
@@ -355,12 +356,16 @@ func (s *anthropicInboundStream) Next() bool {
...
+					toolName = strings.TrimPrefix(toolName, "proxy_")
+
 					s.toolCalls[toolCallIndex] = &llm.ToolCall{
```
looplj, line 358: Seems the stripe will apply for the offical anthropic oauth  type, we can do this in the claude code transformer.

https://github.com/router-for-me/CLIProxyAPI/blob/19b4ef33e06b917450e54171e2b705e34acf1aa3/internal/runtime/executor/claude_executor.go#L127

if isClaudeOAuthToken(apiKey) {
		bodyForUpstream = applyClaudeToolPrefix(body, claudeToolPrefix)
	}

looplj:
Overall looks great. 

I've left a few minor comments for small tweaks/clarifications. Once those are addressed, this should be good to merge.

llm/transformer/anthropic/claudecode/outbound.go
mcowger: Claude Oauth does always uses bearer.   I'll adjust this.

llm/transformer/anthropic/claudecode/utils.go
mcowger: It does not - caching is only impacted by the messages array.   

In my use, I consistently see very very high cache hits (> 90%)

llm/transformer/anthropic/inbound_stream.go
mcowger: Hrm.  I think this is old code from earlier testing.  Will adjust.  Good catch.

llm/transformer/anthropic/claudecode/outbound.go
```diff
@@ -0,0 +1,350 @@
...
+		return nil, fmt.Errorf("failed to get oauth token: %w", err)
+	}
+	apiKey := creds.AccessToken
+
+	// Modify the request body
+	if len(httpReq.Body) > 0 {
+		bodyBytes := httpReq.Body
+
+		// Extract and remove betas array from body
+		extraBetas, bodyBytes := extractAndRemoveBetas(bodyBytes)
```
looplj, line 131: except for the  extractAndRemoveBetas，I think other request body changes can do before   
t.Outbound.TransformRequest(ctx, &reqCopy) to avoid raw []bytes manipulate.

mcowger: Thanks for the feedback!

I've moved `disableThinkingIfToolChoiceForced` to structured pre-processing where it clears `llmRequest.ReasoningEffort` before serialization. However, other manipulations can't move before `t.Outbound.TransformRequest()` because:

1. Anthropic-specific fields (`betas`, system with `cache_control`, `metadata.user_id`) don't exist in the unified `llm.Request` model which is OpenAI-based

2. Tool prefixing would pollute the shared unified model, breaking other transformers that expect original tool names

Moving these would require extending the unified model with Claude Code / Anthropic specific fields or modifying shared state inappropriately. I've documented these architectural constraints in the code. Please let me know if you have suggestions for a cleaner approach, though.

looplj: 1. the llm.Message have the  cache control key to adapter anthropic
https://github.com/looplj/axonhub/blob/f83a732a50d253263d90efdb690b4aa38e417bff/llm/model.go#L330

2. The anthropic transformer will handle the metadata.user_id key， we can inject the fake id to the  metadata user id key
https://github.com/looplj/axonhub/blob/f83a732a50d253263d90efdb690b4aa38e417bff/llm/transformer/anthropic/outbound_convert.go#L42

3. we can applyClaudeToolPrefix before t.Outbound.TransformRequest(ctx, &reqCopy)， because the applyClaudeToolPrefix applied to reqCopy ,will not affect other transformer

mcowger: Good points.  Moved around.

looplj:
Looks great.

May I confirm if the latest code is verified and ready for merge?

mcowger:
It is ready.  I've been using it consistently.  It works flawlessly for me in both claude code client, as well as Kilocode and opencode.

PR #753, opened 2026-02-05
fix: preserve stream flag for TTFT persistence

djdembeck:
## Summary
This PR fixes a bug where TTFT (Time To First Token) latency values were not being persisted and displayed for streaming requests.

### The Problem
In commit 8afd95c3 ("feat: trace sticky api key for multiple api keys channel"), the code was refactored to create a new `PerformanceRecord` instance for each outbound request. While this change was intended to support multiple API keys per channel, it inadvertently broke TTFT tracking.

The issue occurred because:
1. The `Stream` flag was correctly set in `OnInboundLlmRequest()` when processing the initial LLM request
2. However, `OnOutboundRawRequest()` was creating a **brand new** `PerformanceRecord` instance
3. This new instance overwrote the entire struct, including the `Stream` field
4. Without `Stream = true`, the frontend UI gating (`request.stream && metricsFirstTokenLatencyMs`) would fail to display TTFT values

### The Fix
The fix preserves the `Stream` flag from the existing `PerformanceRecord` (if any) before creating the new instance. This ensures that the streaming flag set during inbound request processing is carried through to the performance metrics.

## 问题描述 (Chinese)
此 PR 修复了流式请求的 TTFT（首Token时间）延迟值未被持久化和显示的问题。

### 问题原因
在提交 8afd95c3（"feat: trace sticky api key for multiple api keys channel"）中，代码被重构为为每个出站请求创建新的 `PerformanceRecord` 实例。虽然此更改旨在支持每个渠道的多个 API 密钥，但它无意中破坏了 TTFT 跟踪。

问题发生的原因：
1. `Stream` 标志在处理初始 LLM 请求时在 `OnInboundLlmRequest()` 中被正确设置
2. 但是，`OnOutboundRawRequest()` 正在创建一个**全新的** `PerformanceRecord` 实例
3. 这个新实例覆盖了整个结构，包括 `Stream` 字段
4. 没有 `Stream = true`，前端 UI 条件（`request.stream && metricsFirstTokenLatencyMs`）将无法显示 TTFT 值

### 修复方案
修复方案在创建新实例之前从现有的 `PerformanceRecord`（如果有）中保留 `Stream` 标志。这确保了在入站请求处理期间设置的流式标志被传递到性能指标。

## Root Cause
The root cause is in `internal/server/orchestrator/performance.go`:

**Before (buggy code):**
```go
// Create a new PerformanceRecord instance for each request.
perf := biz.PerformanceRecord{}
perf.StartTime = time.Now()
// ... set other fields
m.outbound.state.Perf = &perf  // This overwrites the Stream flag!
```

**After (fixed code):**
```go
// Preserve Stream flag from existing PerformanceRecord (set in OnInboundLlmRequest)
var streamFlag bool
if m.outbound.state.Perf != nil {
    streamFlag = m.outbound.state.Perf.Stream
}

// Create a new PerformanceRecord instance for each request.
perf := biz.PerformanceRecord{}
perf.StartTime = time.Now()
// ... set other fields
perf.Stream = streamFlag  // Preserve the Stream flag
m.outbound.state.Perf = &perf
```

## Test Coverage
Added comprehensive tests in `internal/server/orchestrator/performance_test.go` to verify:
1. Stream flag is correctly set in `OnInboundLlmRequest()`
2. Stream flag is preserved through `OnOutboundRawRequest()`
3. Non-streaming requests have Stream = false
4. The bug would be caught by the tests if the fix were reverted

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved and propagated the stream preference across request processing so performance tracking consistently respects the original stream setting.

* **Tests**
  * Added comprehensive unit tests validating stream-flag handling across inbound/outbound stages, lifecycle preservation, edge cases, and a regression check to prevent loss of the stream preference.

### DIFF
--- a/internal/server/orchestrator/performance.go
+++ b/internal/server/orchestrator/performance.go
@@ -54,12 +54,19 @@ func (m *performanceRecording) OnOutboundRawRequest(ctx context.Context, request
 		return request, nil
 	}
 
+	// Preserve Stream flag from existing PerformanceRecord (set in OnInboundLlmRequest)
+	var streamFlag bool
+	if m.outbound.state.Perf != nil {
+		streamFlag = m.outbound.state.Perf.Stream
+	}
+
 	// Create a new PerformanceRecord instance for each request.
 	perf := biz.PerformanceRecord{}
 	perf.StartTime = time.Now()
 	perf.ChannelID = channel.ID
 	perf.Success = false
 	perf.RequestCompleted = false
+	perf.Stream = streamFlag
 
 	// Get the API key used for this request from context (set by TraceStickyKeyProvider)
 	if apiKey, ok := contexts.GetChannelAPIKey(ctx); ok {
--- a/internal/server/orchestrator/performance_test.go
+++ b/internal/server/orchestrator/performance_test.go
@@ -1,0 +1,339 @@
+package orchestrator
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/looplj/axonhub/internal/ent"
+	"github.com/looplj/axonhub/internal/server/biz"
+	"github.com/looplj/axonhub/llm"
+	"github.com/looplj/axonhub/llm/httpclient"
+)
+
+// mockChannelService is a mock implementation of ChannelService for testing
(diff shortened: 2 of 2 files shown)


looplj:
Good catch！

Issue #874, opened 2026-02-20
[Bug]: Gemini Vertex AI function calls fail with 'Unknown name id' error


djdembeck:
### Pre-submission Checklist / 提交前检查

- [x] I have searched the existing issues and this bug has not been reported / 我已搜索现有 issues，此 Bug 尚未被报告
- [x] I have read the documentation / 我已阅读文档
- [x] I am using the latest version of AxonHub / 我正在使用 AxonHub 的最新版本

### Bug Description / Bug 描述

When using Gemini models on Google Vertex AI (`gemini_vertex` channel type), function calls (tool calls) fail with the following error: / 在使用 Google Vertex AI (`gemini_vertex` 渠道类型) 上的 Gemini 模型时，函数调用（工具调用）失败，报错如下：

```
failed to stream request: Request failed: Bad Request, error: Invalid JSON payload received. Unknown name "id" at 'contents[1].parts[1].function_call': Cannot find field. Invalid JSON payload received. Unknown name "id" at 'contents[1].parts[1].function_call': Cannot find field. Invalid JSON payload received. Unknown name "id" at 'contents[2].parts[0].function_response': Cannot find field. Invalid JSON payload received. Unknown name "id" at 'contents[2].parts[0].function_response': Cannot find field., code: 400, type: INVALID_ARGUMENT
```

### Root Cause / 根本原因

The issue is in `llm/transformer/gemini/outbound_convert.go`: / 问题出在 `llm/transformer/gemini/outbound_convert.go`：

1. **FunctionCall** (lines 389-394): The `ID` field is unconditionally set on `FunctionCall` objects
2. **FunctionResponse** (lines 447-451): The `ID` field is unconditionally set on `FunctionResponse` objects

**Google Vertex AI's Gemini API does NOT support the `id` field** in `function_call` and `function_response` objects. The standard Gemini API at `generativelanguage.googleapis.com` appears to accept it, but Vertex AI rejects it. / **Google Vertex AI 的 Gemini API 不支持** `function_call` 和 `function_response` 对象中的 `id` 字段。`generativelanguage.googleapis.com` 上的标准 Gemini API 似乎接受它，但 Vertex AI 会拒绝。

### Expected Behavior / 期望行为

Function calls should work on Vertex AI when using Gemini models. / 在 Vertex AI 上使用 Gemini 模型时，函数调用应该正常工作。

### Actual Behavior / 实际行为

[...]

PR #880, opened 2026-02-21
fix(gemini): resolve Vertex AI function call ID and grouping issues (#874)

djdembeck:
See #874

## Summary of Changes

This PR addresses two separate issues with Google Vertex AI Gemini integration:

### 1. Function Call ID Handling (Fixed ✅)
**Problem:** Vertex AI does not support the `id` field in `function_call` and `function_response` objects, causing 400 Bad Request errors.

**Solution:** 
- Keep IDs internally for name lookups
- Clear IDs before JSON serialization for Vertex AI platform only
- See commit `0c5417be`

### 2. Tool Message Grouping (Fixed ✅)
**Problem:** Vertex AI requires that function responses be grouped together in a single Content entry matching the function calls. Previously, each tool message was creating a separate Content entry, causing mismatch errors.

**Solution:**
- Modified `convertLLMToGeminiRequest` to group consecutive tool messages
- Multiple tool results now share a single Content with multiple parts
- See commit `ef8ba05c`

---

## Investigation Notes: STOP Override Issue

During testing, we discovered an **unrelated issue** where Vertex AI Gemini returns `finishReason: "STOP"` prematurely in streaming responses when function calls are involved. We attempted a fix but **reverted it** due to side effects.

### Attempted Solution (Reverted)
We tried tracking tool calls across streaming chunks and requests:
- Added `hasToolCall` to `streamState` to track tool calls across chunks
- Added `lastMessages` storage and `hasUnrespondedToolCalls()` helper to detect cross-request tool calls
- When STOP was received with unresponded tool calls, convert it to `tool_calls`

### Why It Was Reverted
The STOP override caused an edge case where `background_output` would get stuck in a loop. This suggests the state tracking approach interferes with streaming task management. 

**Commit:** `4f2812a2` (added) → `a21814c1` (reverted)

### Future Work
If the premature STOP issue persists, consider:
- Stateless approach: Check request message history directly without storing state
- Per-chunk detection: Look for tool calls in the current response only
- Alternative: Handle at a higher level in the request pipeline

## Summary by CodeRabbit

* **Bug Fixes**
  * Conversion to Gemini requests is now config-aware, improving platform-specific handling.
  * Consecutive tool messages are grouped/merged to produce cleaner Gemini payloads.
  * Vertex AI compatibility: ID fields are automatically removed from function calls/responses.
  * Tool result names are better resolved when identifiers are missing.

* **Tests**
  * Added tests covering Vertex AI ID sanitization and multi-tool message grouping.


looplj (approved):

djdembeck:
Was trying something for an issue I noticed (random interruptions/stops during llm replies, when waiting for a tool call). But ultimately the fixes were a bit too hacky and I didn't want to make this PR any bigger. See description for more info. committed review suggestions

looplj (approved):

PR #915, opened 2026-02-27
fix(orchestrator): detect complete responses without terminal events for Codex + NanoGPT

djdembeck:
## Summary

Fixes an issue where **Codex models (e.g., openai/gpt-5.2-codex) on NanoGPT channels get stuck in "processing" status** even though responses complete successfully.

## Root Cause

Codex models force `Stream=true` but aggregate responses internally, returning complete JSON **without terminal stream events** (like `[DONE]` or `response.completed`). The `InboundPersistentStream` only marked completion when terminal events were detected, so `UpdateRequestCompleted()` was never called.

### `internal/server/orchestrator/inbound.go`
- Added detection logic in `Close()` method (lines 135-146) to handle complete responses without terminal events
- Sets `StreamCompleted=true` when valid aggregated response is detected
- Ensures `UpdateRequestCompleted()` is called for Codex + NanoGPT

### `internal/server/orchestrator/inbound_test.go` (new file)
- Added test for complete response without terminal event
- Added test for terminal event handling (existing behavior)
- Added test for aggregation error path

## Test Results

```
✅ TestInboundPersistentStream_Close_WithCompleteResponse - PASS
✅ TestInboundPersistentStream_Close_WithTerminalEvent - PASS
✅ TestInboundPersistentStream_Close_WithAggregationError - PASS
✅ All existing orchestrator tests - PASS
```

## Docker Image

```bash
docker pull registry.kipp.nz/d2dyno/axonhub:60e3d
```

## Checklist

- [x] Code follows project conventions
- [x] Tests added and passing
- [x] Code reviewed (self-review + tool review)
- [x] Docker image built and pushed
- [x] No breaking changes

Fixes: Codex models stuck in processing on NanoGPT channel

### DIFF
--- a/internal/server/orchestrator/inbound.go
+++ b/internal/server/orchestrator/inbound.go
@@ -140,7 +140,8 @@ func (ts *InboundPersistentStream) Close() error {
 		if err == nil && meta.ID != "" && len(responseBody) > 0 {
 			log.Debug(ctx, "Stream has valid complete response without terminal event, treating as completed")
 			ts.state.StreamCompleted = true
-			ts.persistResponseChunks(ctx)
+			// Use _persistResponse directly to avoid re-aggregating chunks
+			ts._persistResponse(context.WithoutCancel(ctx), responseBody, meta)
 			return ts.stream.Close()
 		}
 	}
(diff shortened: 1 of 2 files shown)


looplj:
Does it work if we can generate a done event for the  NanoGPT channel ?

djdembeck:
@looplj Yes, it works! I tested the NanoGPT channel and all models are generating done events correctly:

- `glm-4.6`: ✅ 1 done event
- `gpt-5.1-codex-mini`: ✅ 1 done event  
- `gpt-5.1-codex-max`: ✅ 1 done event
- `gpt-5.2-codex`: ✅ 1 done event

The original issue was specifically noticed with Codex models getting stuck in "processing" status. The NanoGPT transformer is already handling done events properly for regular chat completions, so this PR adds a safeguard in the orchestrator for edge cases (possibly related to the responses endpoint or certain aggregation scenarios with Codex models).

Let me know if you'd like me to test anything else! 🙏

internal/server/orchestrator/inbound.go
looplj: This line(152) will persistResponseChunks if no error happened, it does not affected by the  StreamCompleted flag.

It should work as expected before add the new  block (line 137-147).

looplj:
The code looks good.

I'am trying to figure out the rood cause.

I added a comment, please help to confirm if I missed something. In in opinion, the code before can handle the case correctly.

djdembeck:
I spent some time trying to reproduce this and found exactly why it was failing, and why my fix handles it!

You are completely right that the normal flow (line 152) handles completion, and the error flow (line 116) correctly marks client disconnects as 'canceled'. 

However, the root cause with Codex models was: **The upstream NanoGPT connection was hanging or dropping abruptly after sending the full JSON, but without sending the `[DONE]` event.**

When this happens:
1. AxonHub receives the full response but hangs waiting for more chunks.
2. The client eventually times out or disconnects (`ctxErr != nil`).
3. `Close()` is finally called. Because `StreamCompleted` is false and `ctxErr` is present, it hits the error block.
4. It marks the request as `canceled` and returns early, entirely skipping line 152!

My PR fixes this by acting as a **global circuit-breaker**. Because `InboundPersistentStream` wraps all channels, it says: "Even if the connection dropped and we didn't get a `[DONE]` event, let's check if the chunks we _did_ receive form a valid, complete response. If they do, save it!"

This isn't just a NanoGPT fix—it protects AxonHub against *any* upstream provider (OpenAI, Anthropic, proxies) that drops a connection after sending valid data but before sending the terminal event.

I just pushed a refactor that cleans up the logic: it now aggregates exactly once (avoiding any double-aggregation penalty) and runs *before* the error check so it correctly catches these client disconnect races. Let me know what you think!

looplj (approved):

looplj:
Thanks for the clarity, the root cause is clear for me .

PR #931, opened 2026-03-01
opt: signature compatible when switch channel in the same session

looplj:


looplj:
@gemini-code-assist  review

PR #941, opened 2026-03-02
fix: cap performance chart display to prevent outlier skew

djdembeck:
## Problem

Dashboard performance charts were being rendered unusable by extreme outlier values. For example, when a model reported 171,000 tokens/second (likely due to near-zero latency measurements), this would completely skew the Y-axis scale, making normal performance values (40-100 TPS) appear as flat lines near zero. Users could not visually compare model performance because the chart scale was dominated by unrealistic outliers.

## Solution

This PR implements a **display-layer capping strategy** that:

1. **Preserves data integrity**: Backend continues to return actual calculated values
2. **Caps chart display only**: Frontend limits visual representation to realistic ranges (1000 TPS / 60s TTFT)
3. **Reveals true values on interaction**: Hover tooltips show actual uncapped values
4. **Standardizes minimum latency**: Enforces 10ms floor across frontend and backend to prevent future extreme calculations

### Backend
- Returns actual throughput/TTFT values without capping
- Enforces 10ms minimum latency constant to prevent artificially high TPS calculations

### Frontend
- Capped chart display with dual data keys (capped for rendering, actual for tooltips)
- Fixed Y-axis tick generation to properly show scale maximum
- Legend buttons display actual averages (not capped values)

## Result

- Charts display readable, comparable performance data
- Users can still access true metrics via hover tooltips
- Prevents future outliers through standardized latency minimums

### DIFF
--- a/internal/server/orchestrator/inbound.go
+++ b/internal/server/orchestrator/inbound.go
@@ -194,6 +194,10 @@ func (ts *InboundPersistentStream) _persistResponse(ctx context.Context, respons
 	if ts.perf != nil {
 		firstTokenLatencyMs, requestLatencyMs, _ := ts.perf.Calculate()
 
+		// Enforce minimum latency to prevent extreme TPS calculations
+		requestLatencyMs = biz.ClampLatency(requestLatencyMs)
+		firstTokenLatencyMs = biz.ClampLatency(firstTokenLatencyMs)
+
 		metrics = &biz.LatencyMetrics{
 			LatencyMs: &requestLatencyMs,
 		}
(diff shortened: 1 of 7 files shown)


internal/server/orchestrator/inbound.go
```diff
@@ -194,6 +194,10 @@ func (ts *InboundPersistentStream) _persistResponse(ctx context.Context, respons
 	if ts.perf != nil {
 		firstTokenLatencyMs, requestLatencyMs, _ := ts.perf.Calculate()
 
+		// Enforce minimum latency to prevent extreme TPS calculations
+		requestLatencyMs = biz.ClampLatency(requestLatencyMs)
```
looplj: we can move the logic to Calculate internal to avoid repeat and easy to missing

djdembeck: @looplj should be all set now :+1:

looplj (approved):

looplj:
All the changes not related to agent will merge to release/v0.9.x branch.

The unstable branch may be unstable for a while, will use release/v0.9.x to release bugfix

PR #945, opened 2026-03-03
feat(dashboard): add all-time token statistics

djdembeck:
This PR adds all-time token statistics to the dashboard, showing cumulative token usage across the entire history of the system. The all-time stats are displayed alongside existing time-range filters (day, week, month) in the Token Stats card.

Screenshot from 2026-03-03 13-13-48

### All-Time Token Statistics
- **New "All" time range tab** in the Token Stats card showing cumulative statistics
- **Real-time aggregation** from usage logs for accurate totals
- **Stale-while-revalidate caching** for optimal performance without sacrificing freshness

### Data Freshness Indicator
- **Info icon with tooltip** (desktop) and **popover** (mobile) showing when all-time stats were last updated
- **Reserved space** prevents layout shift when the icon appears/disappears
- **44x44px touch targets** on mobile for accessibility compliance

### Backend Implementation
- **Stale-while-revalidate caching** with configurable TTL values:
  - Soft TTL (default: 1 hour): Serves stale data while refreshing asynchronously
  - Hard TTL (default: 24 hours): Forces synchronous refresh when exceeded
- **Automatic cache invalidation** when new usage logs are created
- **UTC timestamps** for consistent timezone handling
- **Singleflight pattern** prevents cache stampede under load
- **Thread-safe operations** with proper mutex locking

### Configuration
New optional `dashboard` section in `config.yml`:
```yaml
dashboard:
  all_time_token_stats_soft_ttl: 1h
  all_time_token_stats_hard_ttl: 24h
```

## Technical Details
- Aggregates data from `usage_logs` table for complete historical accuracy
- Cache is invalidated via callback when `UsageLogService` creates new logs
- Uses atomic operations and singleflight for concurrent request handling
- All timestamps use UTC to avoid timezone discrepancies

## UI/UX Improvements
- Added `IconInfoCircle` from @tabler/icons-react for the freshness indicator
- Responsive design with Tooltip for desktop and Popover for mobile
- Fixed-size containers prevent layout shifts when switching time ranges

### DIFF
--- a/frontend/src/features/dashboard/components/token-stats-card.tsx
+++ b/frontend/src/features/dashboard/components/token-stats-card.tsx
@@ -1,13 +1,29 @@
 import { useState } from 'react';
 import { BarChart4 } from 'lucide-react';
+import { IconInfoCircle } from '@tabler/icons-react';
 import { useTranslation } from 'react-i18next';
 import { formatNumber } from '@/utils/format-number';
 import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
 import { Skeleton } from '@/components/ui/skeleton';
 import { Tabs, TabsList, TabsTrigger } from '@/components/ui/tabs';
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from '@/components/ui/tooltip';
+import { Popover, PopoverContent, PopoverTrigger } from '@/components/ui/popover';
 import { useTokenStats } from '../data/dashboard';
 
 type TimeRange = 'allTime' | 'thisMonth' | 'thisWeek' | 'thisDay';
+
+function formatLastUpdated(timestamp: string | null): string {
+  if (!timestamp) return '';
+  const date = new Date(timestamp);
+  return date.toLocaleString('en-US', {
+    month: 'short',
+    day: 'numeric',
+    year: 'numeric',
+    hour: 'numeric',
+    minute: '2-digit',
+    hour12: true,
+  });
+}
 
 export function TokenStatsCard() {
   const { t } = useTranslation();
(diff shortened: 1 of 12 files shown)


internal/server/config.go
```diff
@@ -19,12 +19,26 @@ type Config struct {
 	// LLMRequestTimeout is the maximum duration for processing a request to LLM.
 	LLMRequestTimeout time.Duration `conf:"llm_request_timeout" yaml:"llm_request_timeout" json:"llm_request_timeout"`
 
-	Trace tracing.Config `conf:"trace" yaml:"trace" json:"trace"`
```
looplj, line 22: We should set default values in the conf package

frontend/src/features/dashboard/components/token-stats-card.tsx
```diff
@@ -1,16 +1,76 @@
...
 import { useTranslation } from 'react-i18next';
 import { formatNumber } from '@/utils/format-number';
 import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
 import { Skeleton } from '@/components/ui/skeleton';
 import { Tabs, TabsList, TabsTrigger } from '@/components/ui/tabs';
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from '@/components/ui/tooltip';
+import { Popover, PopoverContent, PopoverTrigger } from '@/components/ui/popover';
 import { useTokenStats } from '../data/dashboard';
 
-type TimeRange = 'thisMonth' | 'thisWeek' | 'thisDay';
```
looplj, line 10: We should show thisDay by default. And maybe we can change the allTime to lazy loading when use switch to allTime

djdembeck: That behaviour is unchanged (thisDay remains default):
https://github.com/looplj/axonhub/pull/945/changes/BASE..c41da589d5227f939f9e89784768541921913a58#diff-6959369d5fdc2c2f7935eab652623b80f88b887cdbca8b8cbad88233e862416eR75

internal/server/config.go
djdembeck: Done 👍

PR #979, opened 2026-03-07
fix: unnecessary exposure of DB port

PA733:
There's no need to expose PG's port, cuz Axonhub can connect to PG via its internal network. The same for MySQL.
Exposure by default may introduce high risks of data lose (especially with default passwd of `axonhub_password`).
Please consider merge this patch ASAP.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,8 +22,6 @@ services:
       POSTGRES_PASSWORD: ${DB_PASSWORD:-axonhub_password}
     volumes:
       - postgres_data:/var/lib/postgresql/data
-    ports:
-      - "5432:5432"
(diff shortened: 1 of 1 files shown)


PA733:
> ## Code Review
> This pull request correctly removes the exposed port for the PostgreSQL service in `docker-compose.yml`, which is an important security improvement as services can communicate over the internal Docker network. My review suggests applying this security principle consistently to the MySQL example configuration within the same file to ensure all provided examples are secure by default.
> 
> I am having trouble creating individual review comments. Click here to see my feedback.

You're right, the same patch applied for MySQL servvice.

looplj (approved):

PR #997, opened 2026-03-08
Handle unsupported metrics exporter types

utafrali:
When metrics are enabled with an unsupported exporter type, the code would panic because the switch statement was missing a default case. Added proper error handling for that scenario, plus tests for disabled metrics and unsupported exporter types.


looplj:
Thanks for the PR. 

The unstable branch is for the v1.0, it will be unstable for a while, we use the release/v0.9.x to release the fix. Please pr based the on the  branch release/v0.9.x .

BTW the metrics are very imperfect for now, maybe we will improve it in the future.

PR #1043, opened 2026-03-12
fix(gc): batch deletes to reduce WAL/lock pressure

doubaoyui:
## What
Reduce GC cleanup WAL/lock spikes by deleting old rows in bounded batches.

## How
- Query only IDs for each batch (avoid loading full rows).
- Delete by ID batch and use affected rows for accounting/logging.
- Guard against a non-empty selected batch deleting 0 rows to avoid a busy loop.

## Verification
- go test ./internal/server/gc


looplj:
可以复用吗，而不是每个地方写一个循环。

PR #1052, opened 2026-03-13
fix: preserve JSON-like string values in header override

GhostFlying:
## Problem
When setting header override values with JSON-like strings (e.g., `{"session_id":"843634473"}`), the previous implementation used `fmt.Sprintf("%v", renderedValue)` which incorrectly formatted the value as a Go map literal with `map[` prefix (e.g., `map[session_id:843634473]`), corrupting the original JSON format.
## Solution
- Replace `renderOverrideValue` with `renderTemplate` to ensure proper string rendering
- Remove `fmt.Sprintf("%v", ...)` formatting and pass the rendered value directly to preserve the original JSON string format
## Changes
- `internal/server/orchestrator/override.go`: Fixed header value assignment in `applyOverrideOperationToHeaders`
- `internal/server/orchestrator/override_test.go`: Added `TestOverrideHeadersKeepJSONLikeString` to verify JSON-like strings are preserved without `map[` prefix
## Test Coverage
New test validates preservation of:
- Mixed-case JSON keys (camelCase)
- Underscore and hyphenated keys
- Complete JSON structure without `map[` prefix contamination

### DIFF
--- a/internal/server/orchestrator/override.go
+++ b/internal/server/orchestrator/override.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"fmt"
 	"net/http"
 	"strings"
 	"text/template"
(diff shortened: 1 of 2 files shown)


looplj:
ci 挂了

GhostFlying:
我修下

looplj (approved):

PR #1059, opened 2026-03-14
fix(orchestrator): preserve completed responses on client disconnect

Chengxiwei:
## What this fixes

When the client disconnects near the end of a streamed response, AxonHub may observe `ctx` cancellation before the local stream consumer sees the terminal event.

In the `Claude Code -> AxonHub -> Codex` chain, that can misclassify an actually completed response as canceled / failed even though the full response has already been streamed upstream.

## Approach

This update keeps the completion proof generic instead of special-casing OpenAI Responses status values.

The outbound finalization logic now treats a response as completed when either:

- an explicit terminal event was observed
- or aggregation succeeds and `meta.Usage != nil`

At the same time:

- explicit upstream stream errors still remain failures
- `ctx` cancellation alone is no longer treated as automatic failure if completion has already been proven
- missing usage is not treated as proof of failure; it is just absence of one completion signal

## Why this shape

The maintainer feedback was to avoid protocol-specific completion handling in the orchestrator.

Using aggregated usage as an additional positive completion signal makes the rule work across providers / formats that already populate usage on complete responses, while still preserving terminal-event-based completion for streams that do not rely on usage.

## Tests

This PR includes:

- targeted orchestrator regression coverage in `internal/server/orchestrator/outbound_test.go`
- an integration-style client-disconnect regression in `integration_test/openai/responses/streaming/client_disconnect_test.go`

Covered cases include:

- missing usage does not count as aggregated completion
- aggregated usage counts as completion
- client cancel after the response is effectively complete is still persisted as completed
- explicit stream errors still remain failures

## Verification

- `go test ./internal/server/orchestrator/...`

### DIFF
--- a/internal/server/orchestrator/outbound.go
+++ b/internal/server/orchestrator/outbound.go
@@ -106,6 +106,7 @@ func (ts *OutboundPersistentStream) Close() error {
 	// even if there's a context cancellation error. This handles the case where
 	// the client disconnects immediately after receiving the last chunk.
 	if ts.state.StreamCompleted {
+		ts.logFinalizationDecision(ctx, "terminal_event_completed", streamErr, ctxErr, true, nil)
 		// Stream completed successfully - perform final persistence
 		log.Debug(ctx, "Stream completed successfully (received [DONE]), performing final persistence")
 		ts.persistResponseChunks(ctx)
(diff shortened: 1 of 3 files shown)


looplj:
应该可以做的通用一点，不需要针对 response 特殊处理。

 agg 以后，如果 usage 有数据，应该就是完整的了

Chengxiwei:
Updated this based on your suggestion.

The completion check is now generic instead of special-casing Responses:
- explicit terminal events still count as completion
- after aggregation, `meta.Usage != nil` is treated as an additional positive completion signal
- explicit upstream stream errors still remain failures

So `ctx` cancellation by itself no longer forces a failed / canceled outcome if completion has already been proven.

I also cleaned up the earlier review leftovers around the removed status-based path and kept the regression coverage for the client-disconnect case.

internal/server/orchestrator/outbound.go
```diff
@@ -106,25 +107,79 @@ func (ts *OutboundPersistentStream) Close() error {
...
+			}
+		}
+
+		return ts.stream.Close()
+	}
+
+	var responseBody []byte
+	var meta llm.ResponseMeta
+	var aggErr error
+	aggregatedStatus := ""
```
looplj: 这个应该不需要了

Chengxiwei: 已删掉，这里确实不需要单独保留 `aggregatedStatus`。现在只保留通用的 `aggregatedCompleted` 判定和相关决策日志。

internal/server/orchestrator/outbound.go
```diff
@@ -136,12 +191,61 @@ func (ts *OutboundPersistentStream) Close() error {
...
+		decision = "completed_via_chunk_persistence"
+	}
+	ts.logFinalizationDecision(ctx, decision, streamErr, ctxErr, aggregatedStatus, aggregatedCompleted, aggErr)
 
-	ts.persistResponseChunks(ctx)
+	if len(responseBody) > 0 {
+		ts.persistAggregatedResponse(context.WithoutCancel(ctx), responseBody, meta)
+	} else {
+		ts.persistResponseChunks(ctx)
+	}
```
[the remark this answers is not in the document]
Chengxiwei: Kept the fallback here on purpose. `aggregatedCompleted` only means aggregation produced enough completion proof (currently via `meta.Usage != nil`); it does not strictly guarantee that every transformer will also return a non-empty `responseBody`. So I kept the `persistResponseChunks` path as a defensive fallback instead of collapsing this block to `persistAggregatedResponse` only.

looplj (approved):

PR #1211, opened 2026-03-28
fix(nanogpt): implement XML tool call parsing for NanoGPT transformer

djdembeck:
## Summary

This PR adds XML tool call parsing support for the NanoGPT transformer to handle multiple malformed XML formats that AI models output, converting them to OpenAI-compatible `tool_calls`.

## Problem

NanoGPT models output tool calls in various inconsistent XML formats instead of the standard OpenAI format:

```xml
<!-- Standard format (rarely seen) -->
<use_tool name="Write"><file_path>X</file_path><content>Y</content></use_tool>

<!-- Malformed formats (common) -->
<Write file_path="X" content="Y"/>
<Write>{"file_path":"X","content":"Y"}</use_tool>
<Write file_path="X">content</use_tool>
<Write><file_path>X</file_path><content>Y</content></Write>
<Write attr="X"
}
</use_tool>  <!-- unclosed opening tag -->
```

These formats were not being parsed, causing test failures and missing tool calls.

## Solution

Implemented a regex-based XML parser in `llm/transformer/nanogpt/xml_parser.go` that:

1. **Pre-checks** content for XML tool call indicators (`MaybeHasXMLToolCalls`)
2. **Normalizes** common malformations (`normalizeXML`):
   - Unclosed opening tags: `<Write attr="..."\n` → `<Write attr="...">\n`
   - Mismatched closing tags: `</Write_file>` → `</Write>`
   - Wrong closing: `</use_use>` → `</use_tool>`
   - Mismatched pairs: `<Write>content</use_tool>` → `<Write>content</Write>`

3. **Parses** multiple XML formats:
   - Self-closing: `<Write file_path="X" content="Y"/>`
   - Simple content: `<Write file_path="X">content</Write>`
   - JSON in content: `<Write>{"file_path":"X"}</use_tool>`
   - Nested elements: `<Write><file_path>X</file_path><content>Y</content></Write>`
   - No space after tag: `<Write_File>{...}</Write_File>`

4. **Extracts** tool name and arguments, converting to OpenAI format

## Changes

- `llm/transformer/nanogpt/model.go` (+17 lines): Add XML parsing call in `ToOpenAIMessage()`
- `llm/transformer/nanogpt/xml_parser.go` (+356 lines): New regex-based XML parser

## Key Features

- ReDoS-safe regex patterns (use `[^<]` instead of `[\s\S]*?` where possible)
- Handles 6+ different XML format variations from NanoGPT
- Escapes quotes in synthetic attributes to prevent XML injection
- Package-level regex compilation for performance
- Maintains existing test compatibility

## Out of Scope

This PR intentionally does NOT handle the following issues, as they are considered model/provider bugs rather than parsing issues:

1. **Tool name concatenation**: When the model generates malformed tag names like `<globglobglobls>` or `<Write_file_path=...>`, these are model output errors. The parser correctly rejects these as unknown tool names.

2. **Missing closing tags in content**: If the model generates `<Write>content` without any closing tag, this is unrecoverable malformed XML. The parser handles some common cases (like unclosed opening tags) but cannot guess missing closing tags.

3. **Nested tool calls**: The parser does not support tool calls nested inside other tool calls (e.g., `<Write><use_tool>...</use_tool></Write>`). This pattern has not been observed in production.

4. **CDATA sections**: XML CDATA sections (`<![CDATA[...]]>`) are not supported as they have not been observed in model outputs.

5. **XML namespaces**: Namespaced tags like `<tools:Write>` are not supported as they have not been observed.

The parser is designed to be lenient for common malformations but not attempt to fix fundamentally broken model output. Errors in model generation should be reported upstream to the model provider.

### DIFF
--- a/llm/transformer/nanogpt/model.go
+++ b/llm/transformer/nanogpt/model.go
@@ -57,10 +57,27 @@ type Message struct {
 
 // ToOpenAIMessage converts the NanoGPT Message to an OpenAI Message.
 // It maps the Reasoning field to ReasoningContent for compatibility.
+// It also parses XML tool calls from content if present.
 func (m *Message) ToOpenAIMessage() openai.Message {
 	// Map reasoning to reasoning_content if present
 	if m.Reasoning != nil {
 		m.ReasoningContent = m.Reasoning
+	}
+
+	// Parse XML tool calls from content if present
+	if m.Content.Content != nil && *m.Content.Content != "" {
+		content := *m.Content.Content
+		if MaybeHasXMLToolCalls(content) {
+			tools, remaining, err := ParseXMLToolCalls(content)
+			if err == nil && len(tools) > 0 {
+				m.ToolCalls = ToOpenAIToolCalls(tools)
+				if remaining != "" {
+					m.Content = ToOpenAIMessageContent(remaining)
+				} else {
+					m.Content = openai.MessageContent{Content: nil}
+				}
+			}
+		}
 	}
 
 	return m.Message
--- a/llm/transformer/nanogpt/xml_parser.go
+++ b/llm/transformer/nanogpt/xml_parser.go
@@ -4,6 +4,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"regexp"
 	"sort"
 	"strings"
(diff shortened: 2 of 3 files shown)


djdembeck:
Had to re-work this multiple times to find a solution that *largely* works to my liking.

Basically, NanoGPT proxies to multiple providers in a round-robin like way. Each provider may have different quants, so model quality (and tool call) can vary significantly, with many GLM 5 models using XML tool calling.

The PR description notes sometimes there are just fully malformed tool calls, which would be a cat and mouse game to fix in axonhub, so I kept the implementation focused on just converting XML tool calls to native tool calls.

looplj (approved):

looplj:
Thanks for the great idea. 
In fact, the universal XML tool call middleware was my original initial idea. However, since most models now have fairly good support for native tool calls, I haven’t gone ahead with implementing it.

PR #1204, opened 2026-03-28
feat: add user agent pass-through settings

djdembeck:
This PR adds user agent pass-through settings for channels.

PR #1231, opened 2026-03-31
feat: add user agent pass-through settings

djdembeck:
Clean implementation of user agent pass-through settings for channels.

Replaces #1204 which had Git contamination issues.

## Changes
- Add user agent pass-through configuration options
- Allow channels to forward original client user agent to upstream providers


djdembeck:
tested and working. Tried combinations of channel-specific setting on/off+global setting on/off.

looplj:
Just confirm, the system will pass-through user-agent by default.

Will this pr break  the default behavior.

djdembeck:
> Just confirm, the system will pass-through user-agent by default.
> 
> Will this pr break the default behavior.

No, the PR keeps the default behavior unchanged: By default, the axonhub user-agent is sent upstream to API endpoints, and the user has to enable user-agent pass-through.

PR #1245, opened 2026-04-02
fix: respect User-Agent header overrides in pass-through middleware

djdembeck:
Fixes an issue where setting User-Agent via channel Override Headers was being overwritten by the pass-through middleware when pass-through was disabled.

**Problem:**
When a channel had User-Agent set via Override Headers with operation "Set" and pass-through was disabled, the User-Agent was being overwritten with the default "axonhub/1.0" value.

**Changes:**
- Add  to detect all User-Agent affecting operations (set, delete, rename, copy)
- Skip pass-through logic when User-Agent override exists
- Add comprehensive tests for all operation types and edge cases

**Testing:**
- All 19 User-Agent pass-through tests pass
- Full orchestrator test suite passes


looplj (approved):

looplj:
maybe we can move the pass-through middleware before apply override

looplj:
sorry，should be move the pass-through middleware before apply override headers.

then the override headers will work as expected

djdembeck:
> sorry，should be move the pass-through middleware before apply override headers.
> 
> then the override headers will work as expected

You are absolutely right. It was a simple ordering issue, updated to be a minimal fix now :+1:

PR #1263, opened 2026-04-03
fix: preserve final prompt tokens in anthropic stream usage

KtzeAbyss:
## Summary
Fix Anthropic streaming usage merge logic so `message_delta` prompt token usage is not overwritten by `message_start` when the final chunk already contains a non-zero value.

## Problem
For Zhipu Anthropic-compatible streaming responses, `message_start.usage.input_tokens` may be `0`, while the final `message_delta.usage.input_tokens` contains the correct value.

AxonHub was always overwriting the final prompt token count with the value saved from `message_start`, which caused the final SSE usage to incorrectly report `input_tokens: 0`.

## Fix
Only fall back to the prompt token count from `message_start` when the final `message_delta` prompt token count is missing or zero.

## Verification
Reproduced locally by comparing `claude-code-hub` and local `axonhub`.

Before the fix:
- AxonHub final `message_delta.usage.input_tokens = 0`

After the fix:
- AxonHub final `message_delta.usage.input_tokens = 10`
- matches `claude-code-hub`

### DIFF
--- a/llm/transformer/anthropic/outbound_stream.go
+++ b/llm/transformer/anthropic/outbound_stream.go
@@ -246,13 +246,15 @@ func (s *outboundStream) transformStreamChunk(event *httpclient.StreamEvent) (*l
 		if streamEvent.Usage != nil {
 			usage := convertToLlmUsage(streamEvent.Usage, state.platformType)
 			if state.streamUsage != nil {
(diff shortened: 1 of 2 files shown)


looplj:
测试挂了

looplj:
不太对，测试数据应该是正确的。
需要看代码哪里有问题

PR #1286, opened 2026-04-04
refactor: add staleness-based cleanup for stuck processing records

djdembeck:
## Summary

Replaces the fingerprint-based approach with a simpler staleness threshold (1 hour) for cleaning up stuck processing records after crashes.

## Key Changes

- **Simplified approach**: Removed server fingerprint tracking entirely
- **Staleness threshold**: Records >1 hour old in processing state are canceled on startup
- **Self-documenting code**: Clear naming (, )
- **Production ready**: Count logging for observability

## Design Rationale

The 1-hour threshold is conservative—no legitimate LLM request should take this long. This approach:
- Eliminates multi-node coordination complexity
- Avoids fingerprint mismatches after restarts
- Recovers from crashes without manual intervention

## Pivot from Original Approach

**Original plan**: Add server fingerprint tracking with shutdown-time cleanup
**Problem**: Fingerprint changes on restart (PID portion), causing stale records to never match
**Solution**: Use staleness threshold only—no fingerprint needed

This is simpler, more reliable, and handles all crash scenarios correctly.


djdembeck:
Changed the scope quite a bit due to gemini review concerns about multi-node instance handling. Some of the diff size is from generated files. Open to input on the node handling part especially!

internal/server/biz/request.go
```diff
@@ -23,6 +25,18 @@ import (
 	"github.com/looplj/axonhub/llm/httpclient"
 )
 
+// getNodeID returns a unique identifier for this node instance.
+// This is used for multi-node safe cleanup of processing records.
+func getNodeID() string {
```
looplj: It is recommended to add a hash for it.

As for node_id, it can easily be confused with the node ID used in GraphQL.

Perhaps we could use server_fingerprint instead.

djdembeck:
## Design Pivot

**Original approach**: Server fingerprint tracking with shutdown-time cleanup
- Added  → renamed to  with hash
- Implemented shutdown hooks to clear owned records
- Complex multi-node coordination

**Problem discovered**: Fingerprint changes on restart (PID portion changes), so stale records from previous runs never matched the new fingerprint. This made the shutdown cleanup ineffective for crash recovery.

**Simplified solution**: Use staleness threshold only (1 hour)
- No fingerprint tracking needed
- No shutdown cleanup needed
- Single 1-hour threshold on startup handles all cases
- Self-documenting code with clear naming

The staleness approach is simpler, more reliable, and correctly handles all crash scenarios without fingerprint coordination complexity.

looplj (approved):

PR #1340, opened 2026-04-09
feat: unify configured model routing and normalize storage keys

yoke233:
## Summary
- unify configured models with channel models for query and routing
- allow model lookup and candidate selection to work consistently with configured model definitions
- keep returned storage keys slash-separated across platforms while still using OS-native paths for local filesystem writes

## Why
There are two related behavior issues here.

First, configured models and channel models were not treated consistently across query, routing, and candidate selection. That created gaps where a model could be configured in AxonHub but still behave differently depending on whether the code path looked at configured models or channel-native models. This change makes those paths converge so configured model definitions can participate in listing and routing more predictably.

Second, data storage keys are logical identifiers and should stay stable across platforms. On Windows, filesystem writes could convert a logical key like `test/key.txt` into `test\\key.txt` and return the converted value. This makes storage keys OS-dependent. The storage change keeps the externally visible key normalized while still using filesystem-specific paths internally where needed.

Putting these changes together keeps model behavior more consistent at the application layer, and keeps persisted storage keys stable at the storage layer.

## Testing
- `go test ./internal/server/biz -count=1`
- `go test ./internal/server/orchestrator -count=1`

### DIFF
--- a/frontend/src/features/channels/components/channels-model-mapping-dialog.tsx
+++ b/frontend/src/features/channels/components/channels-model-mapping-dialog.tsx
@@ -27,7 +27,7 @@ interface Props {
 }
 
 // 扩展 schema 以包含模型映射的校验规则
-const createModelMappingFormSchema = (supportedModels: string[]) =>
+const createModelMappingFormSchema = () =>
   z.object({
     extraModelPrefix: z.string().optional(),
     modelMappings: z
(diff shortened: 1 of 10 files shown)


looplj:
Thanks for the pr.

Please pr based on the release/v0.9.x branch.

yoke233:
根据 review 建议，我已将这组改动重新基于 `release/v0.9.x` 整理并提交到新的 PR：

https://github.com/looplj/axonhub/pull/1384

这个旧 PR 基于 `unstable`，我先关闭，后续请在新 PR 上 review。谢谢。

yoke233:
已改为基于 release/v0.9.x 的新 PR #1384，关闭此旧 PR。

PR #1376, opened 2026-04-11
feat: add nanogpt_responses channel type

djdembeck:
## Summary

This PR adds a new `nanogpt_responses` channel type to support the OpenAI Responses API endpoint for NanoGPT channels.

- `nanogpt` - Routes to `/chat/completions` using custom transformer (reasoning fields + XML tool call parsing)
- `nanogpt_responses` - Routes to `/responses` endpoint using Responses API transformer

The existing `nanogpt` type is preserved with its custom transformer that handles:
- Custom `reasoning` field mapping to `reasoning_content` for compatibility
- XML tool call parsing from content (e.g., `<Write file_path="X">content</Write>`)

## Spirit/Intent

Enable proper API endpoint routing for NanoGPT channels by adding responses API support, allowing users to choose the appropriate endpoint for their use case while maintaining the custom reasoning and XML parsing features for the chat completions endpoint.

### Backend
- Added `nanogpt_responses` enum value to Ent schema
- Regenerated Ent code with new channel type constant
- Added transformer creation logic in `channel_llm.go`:
  - `TypeNanogpt` creates `nanogpt.OutboundTransformer` (custom reasoning + XML parsing)
  - `TypeNanogptResponses` creates `responses.OutboundTransformer`
- Added comprehensive tests for both channel types

### Frontend
- Updated `ChannelType` enum in schema.ts with new type
- Updated nanogpt provider config to include both channel types
- Added channel config for `nanogpt_responses` with correct API format
- Added localization entries for new type

### Tests
- All nanogpt channel tests pass
- All existing nanogpt transformer tests pass
- Frontend TypeScript compilation passes

## Backward Compatibility

- Existing `nanogpt` channels continue to work with custom transformer
- No breaking changes to existing functionality
- Custom reasoning field handling and XML tool call parsing preserved


djdembeck:
Should be good to go :+1: had to make a few changes to simplify things greatly. Eventually need to add Anthropic/`messages` as well since Nano also has that endpoint, but to keep scope limited gonna just split `completions` and `responses` in this PR.

looplj (approved):

PR #1387, opened 2026-04-13
opt: align codex trace header with latest codex cli

yoke233:
## Summary
- align Codex cache/session semantics with the current Codex client headers
- preserve modern Codex identity headers and pass through Codex-specific request metadata
- support `previous_response_id` passthrough for the Responses API path
image

## Why
Recent Codex clients send cache/session context through `X-Codex-Turn-Metadata.session_id` and related Codex headers instead of relying only on legacy `Session_id`. Without honoring these fields, downstream requests can lose conversation/cache continuity and always report `cache_tokens = 0`.

## Changes
- trace middleware: accept `X-Codex-Turn-Metadata.session_id` as trace/session source while keeping `Session_id` precedence
- codex outbound: pass through
  - `X-Codex-Turn-Metadata`
  - `X-Codex-Window-Id`
  - `X-Client-Request-Id`
  - `X-Codex-Beta-Features`
- responses inbound/outbound: preserve `previous_response_id`
- docs: document `previous_response_id` support
- tests: add coverage for new Codex and Responses compatibility paths

## Verification
- `go test ./internal/server/middleware -run 'TestWithTrace_Codex(HeaderSetsTrace|TurnMetadataSetsTrace|HeaderHasPriorityOverTurnMetadata|TurnMetadataInvalidOrMissingSessionDoesNotSetTrace|SessionMissingDoesNotSetTrace)' -count=1`
- `cd llm && go test ./transformer/openai/codex -run 'TestCodexOutbound_(PassthroughModernCodexHeaders|SessionIDPrecedence|MinimalIdentityHeaders|AllowsInboundIdentityOverrides)' -count=1`
- `cd llm && go test ./transformer/openai/responses -run 'TestInboundTransformer_TransformRequest|TestInboundTransformer_TransformResponse|TestOutboundTransformer_TransformRequest|TestOutboundTransformer_TransformResponse' -count=1`

### DIFF
--- a/docs/en/api-reference/openai-api.md
+++ b/docs/en/api-reference/openai-api.md
@@ -59,8 +59,8 @@ AxonHub provides partial support for the OpenAI Responses API. This API offers a
 **Endpoints:**
 - `POST /v1/responses` - Generate a response
 
-**Limitations:**
-- ❌ `previous_response_id` is **not supported** - conversation history must be managed client-side
+**Capabilities:**
+- ✅ `previous_response_id` is supported for continued Responses conversations
 - ✅ Basic response generation is fully functional
 - ✅ Streaming responses are supported
 
@@ -84,9 +84,10 @@ client := openai.NewClient(
 
 ctx := context.Background()
 
-// Generate a response (previous_response_id not supported)
+// Generate a response (previous_response_id supported)
 params := responses.ResponseNewParams{
     Model: shared.ResponsesModel("gpt-4o"),
+    PreviousResponseID: openai.String("resp_prev_123"),
     Input: responses.ResponseNewParamsInputUnion{
         OfString: openai.String("Hello, how are you?"),
     },
--- a/docs/en/api-reference/openai-api.md
+++ b/docs/en/api-reference/openai-api.md
@@ -60,7 +60,7 @@ AxonHub provides partial support for the OpenAI Responses API. This API offers a
 - `POST /v1/responses` - Generate a response
 
 **Capabilities:**
-- ✅ `previous_response_id` is supported for continued Responses conversations
+- ✅ `previous_response_id` passthrough is supported for continued Responses conversations on the same upstream channel
 - ✅ Basic response generation is fully functional
 - ✅ Streaming responses are supported
 
@@ -84,7 +84,7 @@ client := openai.NewClient(
 
 ctx := context.Background()
 
-// Generate a response (previous_response_id supported)
+// Generate a response (previous_response_id passthrough on the same upstream channel)
 params := responses.ResponseNewParams{
     Model: shared.ResponsesModel("gpt-4o"),
     PreviousResponseID: openai.String("resp_prev_123"),
--- a/internal/server/middleware/trace.go
+++ b/internal/server/middleware/trace.go
@@ -2,7 +2,6 @@ package middleware
 
 import (
 	"bytes"
-	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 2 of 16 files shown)


docs/en/api-reference/openai-api.md
```diff
@@ -59,8 +59,8 @@ AxonHub provides partial support for the OpenAI Responses API. This API offers a
 **Endpoints:**
 - `POST /v1/responses` - Generate a response
 
-**Limitations:**
-- ❌ `previous_response_id` is **not supported** - conversation history must be managed client-side
+**Capabilities:**
+- ✅ `previous_response_id` is supported for continued Responses conversations
```
looplj: The previous_response_id is not that easy, this approach limits the turns has to be using the same channel all the time.

llm/transformer/openai/codex/outbound.go
```diff
@@ -163,8 +177,16 @@ func (t *OutboundTransformer) TransformRequest(ctx context.Context, llmReq *llm.
 		hreq.Headers.Set("User-Agent", rawUserAgent)
 	}
 
+	for _, header := range codexPassthroughHeaders {
```
looplj: FYI, this should not be a fix, the legacy Session_Id  header is enough for cache。

Please check if the header is blocked by the nginx.

internal/server/middleware/trace.go
```diff
@@ -188,12 +189,34 @@ func tryExtractTraceIDFromClaudeCodeRequest(c *gin.Context, config tracing.Confi
...
+type codexTurnMetadata struct {
+	SessionID string `json:"session_id"`
+}
 
 // tryExtractTraceIDFromCodexRequest extracts the trace ID from the Codex session header.
 func tryExtractTraceIDFromCodexRequest(c *gin.Context) string {
 	traceID := strings.TrimSpace(c.GetHeader(codexTraceHeader))
 	if traceID == "" {
-		return ""
+		turnMetadataRaw := strings.TrimSpace(c.GetHeader(codexTurnMetadataHeader))
```
looplj: please help to move the code fragment to the codex package, like the claude code.

yoke233:
Addressed the follow-up review points in `daec756e`:

- moved Codex session/header parsing helpers into `llm/transformer/openai/codex/headers.go`, and reused them from `trace.go` + Codex outbound so the Codex-specific fragment no longer lives in middleware
- kept `Session_id` as the primary source, with `X-Codex-Turn-Metadata.session_id` only as fallback compatibility
- fixed the SSE path so `previous_response_id` is preserved both in stream aggregation and outbound streaming chunks
- narrowed the docs wording to same-upstream-channel passthrough instead of broad support wording

I also re-ran the targeted tests:
- `go test ./internal/server/middleware -run 'TestWithTrace_Codex(HeaderSetsTrace|TurnMetadataSetsTrace|HeaderHasPriorityOverTurnMetadata|TurnMetadataInvalidOrMissingSessionDoesNotSetTrace|SessionMissingDoesNotSetTrace)' -count=1`
- `cd llm && go test ./transformer/openai/codex -run 'TestCodexOutbound_(PassthroughModernCodexHeaders|SessionIDPrecedence|MinimalIdentityHeaders|AllowsInboundIdentityOverrides)' -count=1`
- `cd llm && go test ./transformer/openai/responses -run 'TestAggregateStreamChunks|TestOutboundTransformer_TransformStream_PreservesPreviousResponseID|TestInboundTransformer_TransformRequest|TestInboundTransformer_TransformResponse|TestOutboundTransformer_TransformRequest|TestOutboundTransformer_TransformResponse' -count=1`

On the nginx point: agreed this fallback should not replace fixing proxy/header forwarding if `Session_id` is being stripped upstream; this change is only a compatibility guard for the modern Codex header shape.

PR #1401, opened 2026-04-15
fix(channels): bulk weight adjustment not populating on drag reorder

djdembeck:
## Summary

The bulk channel weight ordering dialog did not properly populate weight fields when channels were dragged to reorder. The original `calculateRelativeWeight` function failed when all weights were equal (interpolation produced no gap), and the dialog assigned fabricated rank numbers to channels that had weight 0 in the backend.

## Spirit/Intent

Fix the drag-to-reorder weight assignment so that:
- Dragging a channel to a new position correctly calculates an appropriate weight for its new spot
- Channels with backend weight 0 (unset) are not assigned fabricated numbers on dialog open
- Equal/tied weights are preserved through save and reload
- Lower weights display at the top of the list, higher weights at the bottom
- What the user sees in the dialog is what gets saved to the backend

## Key Changes

- Replaced broken `calculateRelativeWeight` with `reassignWeightsFromPosition` that uses neighbor-based gap logic and shift-below when gaps are too small
- Preserved weight 0 on dialog load instead of fabricating `index + 1` ranks for all channels
- Removed weight inversion conversion on save/load — dialog values are now saved as-is
- Fixed `normalizeRanks` to not reassign sequential weights (was breaking tied weights)
- Changed sort order to ascending so lower weights appear at the top
- Raised `MAX_WEIGHT` from 100 to 9999 to prevent clamping collisions with many channels

## Risks

- Minor: Weight range expanded from 0–100 to 0–9999, but this is more flexible for large channel lists
- Low: Internal sorting logic changed, but UI behavior now matches user expectations

### DIFF
--- a/frontend/src/features/channels/components/channels-bulk-ordering-dialog.tsx
+++ b/frontend/src/features/channels/components/channels-bulk-ordering-dialog.tsx
@@ -21,32 +21,26 @@ const formatWeight = (value: number) => Math.round(value);
 const clampWeight = (value: number) => formatWeight(Math.min(MAX_WEIGHT, Math.max(MIN_WEIGHT, value)));
 
 // Weight 0 = highest priority (sorts to top), otherwise ascending by weight
-const weightSort = (a: { orderingWeight: number }, b: { orderingWeight: number }) => {
-  if (a.orderingWeight === 0 && b.orderingWeight === 0) return 0;
-  if (a.orderingWeight === 0) return -1;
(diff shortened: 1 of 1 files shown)


looplj:
Does the new MAX_WEIGHT will affect the load balance .

djdembeck:
> Does the new MAX_WEIGHT will affect the load balance .

It would have, but I ended up reverting that, since I think it's reasonable to assume a user won't have 100 channels.

PR #1428, opened 2026-04-20
Display userQuery per trace/thread entry

glassbones:
TL;DR: `firstUserQuery` was a bad UI label. In threads it suggested first-ever history, but we want the prompt shown for the current trace/row.

Why:
- thread views are a list of trace rows, so each row should show that trace’s current prompt
- `firstUserQuery` was misleading for both traces and threads
- `userQuery` is clearer and matches the actual span semantics better

Fix:
- keep the backend GraphQL field for compatibility
- alias it to `userQuery` in the frontend
- update trace/thread tables + cards to use the clearer name

Backend note: the trace resolver now returns the latest user query for the current trace segment, which fixes the multi-message thread case.

Before: 

[Thread]
image

[Trace]
image

### DIFF
--- a/internal/server/biz/trace.go
+++ b/internal/server/biz/trace.go
@@ -212,15 +212,11 @@ func (s *Segment) FirstUserQuery() *string {
 	// Prefer the latest user query in the current segment.
 	// The first request of a trace may carry prior thread context as prefix,
 	// so the current turn's user prompt is typically the last user_query span.
-	var latestQuery *string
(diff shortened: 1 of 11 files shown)


internal/server/biz/trace.go
looplj, line 207: I think we should not change the meaning of this method.

The problem is trace level instead of segment level.

We should change the first user query of the trace.

The GetRootSegment handled this case

PR #1461, opened 2026-04-23
fix(playground): only show thinking in last assistant message

myml:
Previously, reasoning (thinking) blocks were displayed for all assistant
messages in the conversation history. Now they only appear for the most
recent assistant response, matching expected behavior.


looplj:
Please help to remove the tsconfig.tsbuildinfo

myml:
> Please help to remove the tsconfig.tsbuildinfo

deleted

myml:
@looplj

PR #1480, opened 2026-04-24
fix(list_models): respect API key profile filtering in full model data mode

djdembeck:
## Purpose/Goal
Fix list models API key profile filtering when full model info setting is enabled.

## Summary
When `include=full` or `DefaultModelAPIIncludeAll=true`, the list_models endpoint was querying all enabled models from the database without respecting API key profile restrictions. This change ensures that API key profile visibility filtering is applied consistently in both basic and extended (full data) modes.

## What Changed
- **internal/server/api/openai.go**: Refactored ListModels handler to first fetch visible models via ModelService (which respects API key profiles), then conditionally enrich with database data based on needFullData flag
- **internal/server/api/openai_retrieve_test.go**: Added comprehensive test coverage:
  - TestOpenAIHandlers_ListModels_ExtendedModeRespectsAPIKeyProfile: Verifies extended mode only returns models the API key has access to
  - TestOpenAIHandlers_ListModels_ExtendedModeFallsBackToBasicForMissingDBModel: Verifies graceful fallback for channel-only models
  - TestOpenAIHandlers_ListModels_ExtendedModeWithZeroAllowedModelsReturnsEmpty: Verifies empty response when API key has no matching models

## Why This Matters
API key profile restrictions were being bypassed when users requested full model metadata. This fix ensures authorization boundaries are maintained regardless of the include parameter or server-side default settings.

## Spirit/Intent
Purpose/Goal: Maintain security invariants while providing rich model metadata. The fix preserves the existing authorization model (API key profiles) while enabling the extended model data feature. Users with restricted API keys should only see models they're authorized to access, whether in basic or extended mode.

### DIFF
--- a/internal/server/api/openai.go
+++ b/internal/server/api/openai.go
@@ -556,23 +556,30 @@ func (handlers *OpenAIHandlers) ListModels(c *gin.Context) {
 		return
 	}
 
+	if len(visibleModels) == 0 {
+		c.JSON(http.StatusOK, gin.H{
+			"object": "list",
+			"data":   []OpenAIModel{},
+		})
+
+		return
+	}
+
 	if !needFullData {
-		openaiModels = make([]OpenAIModel, 0, len(visibleModels))
-		for _, m := range visibleModels {
-			openaiModels = append(openaiModels, convertModelFacadeToOpenAIModel(m))
-		}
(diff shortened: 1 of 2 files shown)


djdembeck:
confirmed working on my instance.

PR #1481, opened 2026-04-25
feat: array override ops (array_append, array_prepend, array_insert)

paopaoandlingyia:
## Motivation

The current four override operations (`set`, `delete`, `rename`, `copy`) have no way to inject items into an existing JSON array without overwriting it — there is no array prepend / append / insert primitive.

## What's added

Three new body override operations:

| Op | Use |
|---|---|
| `array_append` | insert value(s) at the end of the target array |
| `array_prepend` | insert value(s) at the start of the target array |
| `array_insert` | insert value(s) at a specific position (`op.index`) |

Two new optional fields on `OverrideOperation`:

| Field | Default | Notes |
|---|---|---|
| `index` (`*int`) | required for `array_insert` | negative values count from the end; out-of-range values are clamped to `[0, len]` |
| `splat` (`*bool`) | `true` | when the rendered value is a JSON array, spread its elements into the target; set `false` to insert as a single nested element |

## Behavior

- If `path` doesn't exist, a new array is created with the value(s).
- If `path` exists but isn't an array, the operation is skipped with a warning.
- Multiple array ops on the same path are additive — each runs against the current state.
- Array ops on headers are not supported and will be rejected by the validator.
- Template rendering (`{{.Model}}`, `{{index .Metadata "key"}}` etc.) is supported in `value`, same as `set`.


looplj:
感谢 PR，帮忙确认下是否本地验证过

paopaoandlingyia:
> 感谢 PR，帮忙确认下是否本地验证过

已本地验证过：新增 array ops 单测、biz override 校验相关测试、以及 orchestrator override 相关用例均通过；本地也做过一次构建检查，未发现问题。

PR #1516, opened 2026-04-27
feat: add optional read-write separation for database

lazhenyi:
## Summary                                                                                                                                                                                               
                                         
  Add optional read-write separation for the database. When `read_replica.read_dsn` is configured, read-only queries (SELECT/WITH, etc.) are automatically routed to the replica, while writes and         
  transactions always go to the master. When unconfigured, everything falls back to single-master mode with zero overhead.
                                                                                                                                                                                                           
  ## Changes                                                

  - **db/config.go**: Add `ReadReplicaConfig` (`read_dsn`, `read_max_open_conns`, `read_max_idle_conns`)
  - **db/ent.go**: Refactor `NewEntClient`, extract `openDB()` / `driverName()` / `entDialect()`; enable router driver only when `read_dsn` is non-empty
  - **db/router.go**: New `routerDriver` that inspects the first SQL token for routing + `masterTx` to keep transactions on master
  - **conf/conf.go**: Add three `read_replica` defaults (empty/zero)
  - **config.example.yml / deploy/helm/**: Add config examples and env var comments
  - **docs/**: Add "Read-Write Separation" section to both zh and en configuration docs

  ## Notes

  - Not applicable to SQLite (single file, no replication)
  - Transactions (`client.Tx()`) always go to master to avoid replication lag
  - Only requires `read_dsn` to enable — zero business code changes

PR #1503, opened 2026-04-26
feat: per-channel concurrency queue + admission control

suixinio:
Closes #1130

## Summary

PR #1322 only added scoring-layer down-ranking for `MaxConcurrent`, which was a no-op when a model had only one channel. This PR adds true admission enforcement plus a FIFO queue, observability, and live UI feedback.

**Backend**
- `ChannelLimiter` with two modes: soft (count only — preserves PR #1322 behaviour) and hard (FIFO queue + capacity ceiling + per-channel timeout). FIFO uses `container/list` with slot-handoff `Release` so released slots transfer directly to the head waiter — preserves fairness and avoids the "newcomer steals released slot" race.
- `ChannelLimiterManager`: per-channel limiter cache keyed by `limiterConfig` struct equality (no hashing); exposes `Snapshot()` for observability.
- `ChannelQueueError`: typed sentinel + synthetic `*httpclient.Error` so `errors.AsType[*httpclient.Error]` plucks out a 429 + structured JSON body (`channel_queue_full` / `channel_queue_timeout`). Deliberately no `Retry-After` header so the upstream-cooldown middleware does not misclassify local rejections as upstream 429s.
- New `ChannelRateLimit` fields: `QueueSize`, `QueueTimeoutMs` (both `*int64`, ms units, matching the existing `metrics_*_ms` convention).
- Middleware order: channel admission runs **before** rate-limit tracking so locally rejected requests do not consume RPM budget.
- LB scoring composes RPM/TPM and concurrency sub-scores via `min()`; hard-mode queueing is capped at 30 % so an idle peer keeps a clear advantage.
- Removed the legacy `ConnectionTracker` (and the unused `ConnectionAwareStrategy`); `ChannelLimiterManager` is the single source of truth for in-flight stats.
- OpenTelemetry: 5 channel-level instruments (`axonhub_channel_inflight`, `_queue_waiting` gauges; `_queue_full_total`, `_queue_timeout_total` counters; `_queue_wait_seconds` histogram).
- `ChannelService.ValidateRateLimit` rejects negatives and `queueSize > 0` without `maxConcurrent > 0`.
- New GraphQL field `Channel.liveLimiterStats` returning real-time `{inFlight, waiting, capacity, queueSize}`.

**Frontend**
- Rate-limit dialog: new `Queue Size` and `Queue Wait Timeout` inputs with cross-field validation; amber advisory when only `MaxConcurrent` is set (soft-mode warning).
[...]

PR #1556, opened 2026-05-01
perf(httpclient): persistent goroutine for SSE decoder

hdot123:
## Problem

`Next()` each time creates a new goroutine + channel to wrap the blocking `sseStream.Recv()` call. In high-throughput SSE scenarios (e.g. concurrent streaming LLM responses), this causes significant GC pressure and scheduling overhead.

## Fix

- Start a single persistent `recvLoop` goroutine at decoder initialization
- Use a buffered channel (cap=1) to relay results to `Next()`
- `Close()` cancels context → goroutine exits → channel closed, no leak
- Context cancellation semantics fully preserved

## Scope

Single file change: `llm/httpclient/decoder.go` (plus minimal test field rename for compilation)

## Test

All existing decoder tests pass.

### DIFF
--- a/llm/httpclient/client_test.go
+++ b/llm/httpclient/client_test.go
@@ -602,7 +602,9 @@ data: [DONE]
 	body := io.NopCloser(strings.NewReader(sseData))
 
 	stream := &defaultSSEDecoder{
-		ctx:       t.Context(),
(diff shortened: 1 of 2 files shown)


hdot123:

PR #1557, opened 2026-05-01
fix(bailian): strip response_format and sanitize tool arguments

hdot123:
## Problem

Bailian (百炼) API:
1. Does not support the `response_format` parameter — forwarding it causes errors
2. Requires tool call arguments to be valid JSON — malformed arguments cause API failures

## Fix

- Strip `response_format` from outgoing requests to Bailian
- Sanitize tool call arguments to ensure valid JSON before forwarding
- Add unit tests for sanitization logic

## Scope

`llm/transformer/bailian/outbound.go` + new test file

### DIFF
--- a/llm/transformer/bailian/outbound.go
+++ b/llm/transformer/bailian/outbound.go
@@ -2,6 +2,7 @@ package bailian
 
 import (
 	"context"
+	"encoding/json"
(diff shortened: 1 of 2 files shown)


hdot123:

PR #1560, opened 2026-05-01
feat(auth): add JWT token revocation mechanism

hdot123:
## Problem

JWT tokens have a 7-day lifetime with no way to revoke them individually.
If a token is compromised, it remains valid until natural expiration.

## Fix

- Add `TokenRevocationService` with in-memory revocation table + periodic sweeper
- Add `jti` (JWT ID) claim to generated tokens
- Add `Revoke(jti)` method for explicit token invalidation
- `AuthenticateJWTToken` checks revocation status before accepting
- Extract `TokenTTL` as shared constant

## Design Notes

- Revocation TTL matches TokenTTL, ensuring revoked entries outlive their tokens
- Sweeper runs hourly to clean expired entries
- Thread-safe via sync.Map

## Scope

Single file: `internal/server/biz/auth.go`

### DIFF
--- a/internal/server/biz/auth.go
+++ b/internal/server/biz/auth.go
@@ -6,9 +6,11 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/golang-jwt/jwt/v5"
+	"github.com/google/uuid"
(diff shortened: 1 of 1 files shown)


hdot123:

PR #1561, opened 2026-05-01
fix(migration): add rollback support and fix completed tracking

hdot123:
## Problem

When a data migration fails:
1. No rollback mechanism exists to handle partially completed migration runs
2. The failed version was incorrectly tracked in the completed list

## Fix

- Add `Rollback()` method that reverses successfully completed migrations (in reverse order) for diagnostics
- `completed` list now only contains versions that truly completed successfully
- Failed migrations trigger Rollback of previously successful ones in the same run
- Thread-safe via mutex

## Scope

`internal/ent/migrate/datamigrate/migrator.go`

### DIFF
--- a/internal/ent/migrate/datamigrate/migrator.go
+++ b/internal/ent/migrate/datamigrate/migrator.go
@@ -2,6 +2,8 @@ package datamigrate
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


hdot123:

PR #1562, opened 2026-05-01
fix(httpclient): propagate Close errors and add ResponseHeaderTimeout

hdot123:
## Problem

1. `SaveDataFromReader` uses `defer f.Close()` without checking the error, potentially losing write failures
2. HTTP transport may lack explicit `ResponseHeaderTimeout` causing indefinite hangs on unresponsive servers

## Fix

- Replace `defer f.Close()` with explicit Close + error check
- Ensure `ResponseHeaderTimeout` is set on transport config

## Scope

`llm/httpclient/` package files only

### DIFF
--- a/llm/httpclient/client.go
+++ b/llm/httpclient/client.go
@@ -16,6 +16,12 @@ import (
 	"github.com/looplj/axonhub/llm/streams"
 )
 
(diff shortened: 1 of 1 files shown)


hdot123:

PR #1567, opened 2026-05-01
fix(resilience): sliding window and half-open circuit breaker fixes

hdot123:
## Problem

Circuit breaker has two issues:
- Sliding window counter resets incorrectly, losing failure history
- Half-open state transition doesn't properly allow probe requests through

## Fix

- Fix sliding window counter reset logic to maintain accurate failure counts
- Fix half-open state to correctly transition and allow probe requests

## Scope

Circuit breaker related files only.


hdot123:

PR #1568, opened 2026-05-01
fix(graphql): N+1 query optimization and builder panic guard

hdot123:
## Problem

- GraphQL queries trigger N+1 database queries (one per node in a list)
- GraphQL Nodes resolver has scope issues
- Builder can panic on invalid/nil input

## Fix

- Optimize N+1 queries with batched loading
- Fix Nodes resolver scope
- Add nil guards to prevent builder panics

## Scope

GraphQL and builder related files only.


hdot123:

PR #1573, opened 2026-05-01
fix(bailian): strip unsupported response_format field

hdot123:
## Problem

Bailian API does not support the `response_format` parameter. Forwarding it causes API errors.

## Fix

Strip `ResponseFormat` from outgoing requests in the Bailian transformer before forwarding.

## Scope

Single file: `llm/transformer/bailian/outbound.go`


looplj:
https://bailian.console.aliyun.com/cn-beijing?spm=5176.29619931.J_PvCec88exbQTi-U433Fxg.4.74cd10d7yO9Xyz&tab=api#/api/?type=model&url=3016807

image

PR #1578, opened 2026-05-01
fix: prevent builder panic and implement Nodes resolver

hdot123:
## Problem

1. `RequestBuilder.WithBody()` panics on `json.Marshal` failure — should return error
2. GraphQL `Nodes` resolver is unimplemented (panic placeholder)

## Fix

- Replace panic with error accumulation in `WithBody()`; error returned at `Build()` time
- `Build()` signature changed from `*Request` to `(*Request, error)`
- All callers updated to handle the new error return
- Implement `Nodes` resolver using `client.Noders()` for batch loading

## Scope

`llm/httpclient/builder.go` + `internal/server/gql/ent.resolvers.go` + 8 caller files

### DIFF
--- a/internal/server/biz/data_storage.go
+++ b/internal/server/biz/data_storage.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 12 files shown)


looplj:
设计如此。

建议还是看看能不能解决一下 issue 里面的问题。

而不是仅仅用 AI 分析一下，虚空打靶。

PR #1657, opened 2026-05-13
fix(transformer/copilot): strip reasoning input items from Responses API requests

djdembeck:
## Summary

GitHub Copilot's Responses API endpoint rejects requests containing `type="reasoning"` input items with a 400 error: `input[N].type "reasoning" is not a valid input item type`. This happens when multi-turn conversations include prior assistant messages that carried OpenAI-style encrypted reasoning content (e.g. from GPT-5.4 models), which gets forwarded verbatim on subsequent requests.

Note: This is distinct from #1656 which changed the chat-completions `ReasoningField` default. That fix doesn't help here because the Responses API uses a different input schema with typed `Item` objects — the `type="reasoning"` item is a Responses-API-specific construct.

## Spirit/Intent

Prevent 400 errors when proxying multi-turn Responses API requests through providers that don't support reasoning input items.

## Key Changes

- Add `StripInputReasoning` config field to `responses.Config` that removes `type="reasoning"` input items from the outbound request payload
- Thread `stripInputReasoning` flag through `convertInputFromMessages` → `convertAssistantMessage` call chain
- Enable `StripInputReasoning: true` in the Copilot outbound transformer (the primary provider that rejects these items)
- Add comprehensive tests covering: reasoning preserved (flag=false), reasoning stripped (flag=true), reasoning+tool-calls stripped, and no-signature passthrough

## Risks

- Providers that **do** support OpenAI reasoning caching (e.g. direct OpenAI API) will lose reasoning context on multi-turn when this flag is enabled. The flag defaults to `false` and is only enabled for Copilot where it's required.
- The encrypted content in reasoning items is opaque — stripping it means the upstream cannot continue a cached reasoning chain, but Copilot doesn't support that mechanism anyway.

### DIFF
--- a/internal/server/biz/channel_endpoint.go
+++ b/internal/server/biz/channel_endpoint.go
@@ -146,7 +146,10 @@ var defaultEndpointsForChannelType = map[channel.Type][]objects.ChannelEndpoint{
 		{APIFormat: llm.APIFormatJinaEmbedding.String()},
 	},
 	channel.TypeGithub:           openAICompatibleDefaultEndpoints,
-	channel.TypeGithubCopilot:    {{APIFormat: llm.APIFormatOpenAIChatCompletion.String()}},
+	channel.TypeGithubCopilot: {
+		{APIFormat: llm.APIFormatOpenAIChatCompletion.String()},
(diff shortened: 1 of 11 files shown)


djdembeck:
investigating another case with this that came up

Issue #1636, opened 2026-05-10
[Bug/错误]: 使用 codex + auth.json 时无法 featch models


xbpk3t:
### Pre-submission Checklist / 提交前检查

- [x] I have searched the [existing issues](https://github.com/looplj/axonhub/issues) and this bug has not been reported / 我已搜索现有 issues，此 Bug 尚未被报告
- [x] I have read the [documentation](https://github.com/looplj/axonhub#readme) / 我已阅读文档
- [x] I am using the latest version of AxonHub / 我正在使用 AxonHub 的最新版本

### FAQ Check / FAQ 检查

- [x] I have checked the [FAQ](https://github.com/looplj/axonhub/issues/1374) and this bug is not covered there / 我已查阅 FAQ，此 Bug 未被覆盖

### Summary

Codex channel created with imported `auth.json` works after manually entering models, but `Fetch Models` fails during channel creation.

### Version

AxonHub: v0.9.38
Deployment: Docker

### Steps to reproduce

1. Add Channel
2. Select Codex
3. Choose `auth.json` auth mode
4. Paste a valid Codex `auth.json`
5. Click `Apply To Credentials`
6. Click `Fetch Models`

### Actual behavior

`Fetch Models` returns:

```json
{
  "models": [],
  "error": "failed to fetch models: GET - https://chatgpt.com/backend-api/codex/models with status 400 Bad Request"
}
````

Browser GraphQL request shows `apiKey` is sent as an access token JWT, not the full OAuth credentials JSON.

### Expected behavior

For Codex official/auth.json credentials, `FetchModels` should return the built-in `codex.DefaultModels()` list instead of requesting:

```text
https://chatgpt.com/backend-api/codex/models
```

### Additional context

If I manually fill the model IDs, the channel works correctly. So the auth.json and actual Codex request path are valid; only the Fetch Models helper fails.

## 可能的修复方向

后端修会更稳。比如：

```go
if channelType == channel.TypeCodex {
    if auth-json mode / official mode / default codex baseURL {
        return codex.DefaultModels()
    }
}
````

或者前端在 auth.json 模式下调用 `FetchModels` 时，不要把 `access_token` 单独传给 `apiKey`，而是传完整 OAuth credentials JSON，让后端的 `isOAuthJSON(apiKey)` 能命中默认模型分支。

---

发 issue 后，再次确认了一下

在第一次新增 auth.json 时会有上面的问题

但是现在又可以成功 fetch models 了

[...]

PR #1665, opened 2026-05-14
fix: add defensive chunking for SQL IN clauses and remove external_id length limit

survivor998:
﻿## Summary

- **Chunk all SQL IN clause calls** with a safe limit of 500 variables to prevent SQLite "too many SQL variables" error (limit is 999 for SQLite < 3.32.0, 32766 for newer versions)
- **Remove MaxLen(512) from external_id fields** and use MySQL text type via SchemaType, preventing truncation of long external IDs (related to #1471)

## Problem

The /models API 500 error was caused by unchunked SQL IN clauses exceeding SQLite variable limit. The original fix in openai.go only protected one call path, leaving several others vulnerable:

1. queryConfiguredModelFacades - ModelIDIn without chunking
2. BulkArchiveModels/BulkDisableModels/BulkEnableModels/BulkDeleteModels - IDIn without chunking
3. BulkCreateModels - complex OR predicate with 2 variables per input

Additionally, external_id field had MaxLen(512) which caused MySQL Data too long errors with long external IDs (e.g. Copilot Response API IDs >400 chars, as reported in #1471).

### Defensive Chunking (biz.SQLiteMaxVariableLimit = 500)
- queryConfiguredModelFacades: chunk ModelIDIn calls
- BulkArchiveModels/BulkDisableModels/BulkEnableModels/BulkDeleteModels: chunk IDIn calls
- BulkCreateModels: chunk existence check (2 vars per predicate, effective chunk size 250)
- openai.go ListModels: replace hardcoded 500 with biz.SQLiteMaxVariableLimit

### Schema Changes
- request.go: external_id from MaxLen(512) to SchemaType(text)
- request_execution.go: same change

### Post-merge Required
- Run go generate ./internal/server/gql to regenerate Ent code after schema changes

## Related Issues
- #1471 (external_id too long for MySQL)
- #1636 (codex fetch models fails)

### DIFF
--- a/internal/ent/schema/request.go
+++ b/internal/ent/schema/request.go
@@ -3,6 +3,7 @@ package schema
 import (
 	"entgo.io/contrib/entgql"
 	"entgo.io/ent"
+	"entgo.io/ent/dialect"
 	"entgo.io/ent/schema"
 	"entgo.io/ent/schema/edge"
 	"entgo.io/ent/schema/field"
@@ -88,7 +89,9 @@ func (Request) Fields() []ent.Field {
 		// External ID for tracking requests in external systems
 		field.String("external_id").
 			Optional().
-			MaxLen(512),
+			SchemaType(map[string]string{
+				dialect.MySQL: "text",
+			}),
 		// The status of the request.
 		field.Enum("status").Values("pending", "processing", "completed", "failed", "canceled"),
 		// Whether the request is a streaming request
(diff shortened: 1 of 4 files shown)


looplj:
Is it a real case for more than 500 models and external id more than 512 chars?

llc1123:
Closing as a duplicate of #1668 (same author; the chunking + external_id changes here are an exact patch-level subset of #1668 — identical patch-id). Let's keep #1668 as the canonical branch for this work. Note the core fix (SQL IN chunking + external_id length) is still pending there; please rebase/split #1668 to isolate it.

PR #1755, opened 2026-05-31
fix(openai): ensure reasoning_content is present in all assistant messages

ifdog:
When reasoning_content is enabled (ReasoningFieldContent or ReasoningFieldAll), ensure all assistant messages have a reasoning_content field.

Some thinking-enabled providers (e.g., Kimi, DeepSeek) require reasoning_content to be present in every assistant message, even if it's an empty string. Without this, multi-round tool-calling conversations fail with:
> thinking is enabled but reasoning_content is missing in assistant tool call message

This aligns with the existing fix already present in the DeepSeek transformer.


looplj:
ds/mimo 是这样，但是不代表其他都是这样吧，不确定是否有其他兼容性问题。
而且，现在 ds 应该已经修复这个问题了才对

PR #1859, opened 2026-06-17
fix(frontend): hide api key creator column without system user permission

KremeCN:
## Summary
- Hide the API key Creator column unless the current user has system-level `read_users` access
- Avoid showing a misleading deleted-user fallback when creator data is not readable

## Problem
The API Keys table always rendered the Creator column, but `APIKey.user` can resolve to `null` when the related user is not readable under the current permission model.

In that case, the frontend fell back to the deleted-user label even though the creator might still exist and was simply not readable.

## Fix
Only add the Creator column when the current user has system-level `read_users` access.

This keeps the UI aligned with the current backend `User` privacy behavior without changing the underlying permission model.

## Test Plan
- [x] Verified by code inspection that the Creator column is now gated by system-level `read_users`
- [x] Verified by code inspection that the Creator column defaults to hidden when the new permission argument is omitted
- [ ] Not run (UI-only change)

### DIFF
--- a/frontend/src/features/apikeys/components/apikeys-columns.tsx
+++ b/frontend/src/features/apikeys/components/apikeys-columns.tsx
@@ -41,7 +41,7 @@ function ApiKeyCell({ apiKey, fullApiKey }: { apiKey: string; fullApiKey: ApiKey
   );
 }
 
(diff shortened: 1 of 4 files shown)


KremeCN:
Addressed the Greptile note in 57bf693e: the Creator filter now uses the same `canViewCreators` visibility condition as the Creator column. @greptileai

PR #1902, opened 2026-06-23
fix(frontend): avoid logout on upstream auth errors

lihaojiee:
## Summary
- Stop treating arbitrary GraphQL error message text containing "Unauthorized" as an admin session-expiration signal.
- Keep logout behavior for explicit HTTP 401/403 responses and GraphQL `UNAUTHENTICATED` extension codes.
- Add a small unit regression test for upstream provider 401 errors returned inside GraphQL business errors.

## Example
A channel model-sync request can return HTTP 200 with a GraphQL business error like:

```json
{
  "errors": [
    {
      "message": "model fetch returned error: failed to fetch models: GET - https://provider.example/v1/models with status 401 Unauthorized",
      "path": ["syncChannelModels"]
    }
  ],
  "data": null
}
```

The `Unauthorized` text here belongs to the upstream provider request, not the AxonHub admin JWT session. The frontend should show it as an operation error instead of clearing the admin token and redirecting to sign-in.

## Test Plan
- `node --test src/**/*.test.mjs` from `frontend/`

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -13,6 +13,7 @@
     "format": "prettier --write .",
     "prepare": "husky",
     "knip": "knip",
+    "test:unit": "node --test src/**/*.test.mjs",
     "test:e2e": "../scripts/e2e/e2e-test.sh",
     "test:e2e:headed": "../scripts/e2e/e2e-test.sh --headed",
     "test:e2e:ui": "../scripts/e2e/e2e-test.sh --ui",
--- a/frontend/src/gql/graphql.test.mjs
+++ b/frontend/src/gql/graphql.test.mjs
@@ -1,0 +1,33 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+import ts from 'typescript';
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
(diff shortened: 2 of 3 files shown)


lihaojiee:
I checked this against the current backend error paths. I don't think this PR should change `isForbiddenGraphQLError` at the same time.

The auth case fixed here is different because a GraphQL message-text match triggers `removeTokenFromStorage()` and a `/sign-in` redirect, so an upstream provider `401 Unauthorized` can destroy the admin session.

For forbidden errors, the current backend still has real permission-denied paths that return plain errors without `extensions.code: FORBIDDEN`, for example:

- `internal/server/gql/system.resolvers.go`: `permission denied: requires write:channels scope`
- `internal/server/gql/me.resolvers.go`: `permission denied: this identity does not belong to you`
- several biz service permission checks that wrap errors as `permission denied: ...`

So changing forbidden detection to extension-code-only in this PR would risk regressing existing permission handling. A cleaner follow-up would be to standardize those backend permission errors to return `xerrors.ForbiddenError(...)`, then remove the frontend message fallback for forbidden errors separately.

PR #1940, opened 2026-06-29
fix: retry stream errors before client commit

lihaojiee:
## Summary
- Add a configurable streaming probe window before committing a stream to the client, so upstream stream errors seen during the window return to the pipeline retry loop.
- Wire the setting through retry policy GraphQL/API/frontend settings with a 0 ms default and 120000 ms clamp.
- Fix Windows test compatibility for embedded OpenAI fake responses and reused WebSocket close errors.

## Example
Request #29292 failed after the upstream stream had been established with stream error: stream ID 7; INTERNAL_ERROR; received from peer. Previously that error only surfaced from stream.Err() in the HTTP handler, after pipeline.Process had already returned, so retry was never attempted. With streamProbeDurationMs configured, the pipeline owns the stream during the probe window and returns retryable upstream errors before any client bytes are committed.

## Test Plan
- cd internal/server/gql && go generate
- go test -p 1 ./...
- cd llm && go test -p 1 ./...
- go build -p 1 -ldflags "-s -w" -tags=nomsgpack -o TEMP/axonhub-pr-check/axonhub.exe ./cmd/axonhub
- cd cmd/schema && go build -p 1 -o TEMP/axonhub-pr-check/schema.exe
- cd llm && go build -p 1 ./...
- go build -p 1 ./...
- cd frontend && CI=true corepack pnpm@10 install --frozen-lockfile
- cd frontend && corepack pnpm@10 build
- go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest run --timeout 10m --max-same-issues 50
- git diff --check
- JSON parse check for frontend/src/locales/en/system.json and frontend/src/locales/zh-CN/system.json

### DIFF
--- a/frontend/src/features/system/components/retry-settings.tsx
+++ b/frontend/src/features/system/components/retry-settings.tsx
@@ -27,6 +27,7 @@ export function RetrySettings() {
     nonStreamResponseTimeoutSeconds: 0,
     loadBalancerStrategy: 'adaptive',
     emptyResponseDetection: false,
+    streamProbeDurationMs: 0,
     upstreamErrorPolicy: {
       mode: 'passthrough',
       customMessage: '',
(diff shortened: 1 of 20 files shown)


lihaojiee:
Closing this PR; reverting the stream retry probe changes for now.

PR #1959, opened 2026-07-03
fix(aggregator): skip non-standard SSE events with empty choices

Black0416:
## Bug

When a provider sends non-standard SSE events with empty `choices: []` after the last valid chat completion chunk but before/after `[DONE]`, two issues occur:

1. **DB aggregation**: `AggregateStreamChunks` unconditionally overwrites `lastChunkResponse` with the non-standard event, causing `id`/`model`/`created` to be zero-valued in the persisted `response_body`.

2. **Client SSE stream**: `TransformStreamChunk` forwards the non-standard event to the client, causing downstream consumers to read `usage: null` as the final usage.

### Reproducible with

**OpenCode Go** (`opencode.ai/zen/go/v1`) which emits:
- `inference-cost` event (before `[DONE]`): `{"choices":[],"x-opencode-type":"inference-cost","cost":"...","normalizedUsage":{...}}`
- `cost` event (after `[DONE]`): `{"choices":[],"cost":"0"}`

Both have `choices: []` and `usage: null`, but `json.Unmarshal` succeeds so they are not skipped.

### Why `usage` was unaffected in DB but affected in client stream

- **DB**: `usage` uses conditional assignment: `if chunk.Usage != nil` so it is not overwritten by nil usage.
- **Client stream**: `TransformStreamChunk` forwards each event as `llm.Response` to the client. The last non-`[DONE]` event has `usage: null`.

## Root Cause

In `llm/transformer/openai/aggregator.go`:
```go
lastChunkResponse = chunk  // unconditionally overwrites
```

In `llm/transformer/openai/outbound.go` `TransformStreamChunk`:
```go
return t.TransformResponse(ctx, httpResp)  // forwards non-standard events as-is
```

### 1. Aggregator - skip chunks with empty choices

```diff
-		lastChunkResponse = chunk
+		if len(chunk.Choices) > 0 {
+			lastChunkResponse = chunk
+		}
```

Standard chat completion chunks always have at least one choice.

### 2. TransformStreamChunk - skip non-standard events in client stream

```diff
+	if len(resp.Choices) == 0 {
+		return nil, nil
+	}
 	return resp, nil
```

Wrap `TransformStream` with `streams.NoNil` to filter out nil responses.

## Impact

- Only affects non-standard events with `choices: []`
- All standard OpenAI/DeepSeek/Anthropic chunk paths unchanged
- Zero risk to existing behavior

## Verification

Tested with OpenCode Go (affected) and JDCloud (unaffected):
- DB `response_body`: `id`/`model`/`created` correctly preserved
- Client SSE stream: non-standard events filtered out
- Downstream token tracking: input/output/cache all aligned

### DIFF
--- a/llm/transformer/openai/aggregator.go
+++ b/llm/transformer/openai/aggregator.go
@@ -246,8 +246,12 @@ func AggregateStreamChunks(ctx context.Context, chunks []*httpclient.StreamEvent
 			systemFingerprint = chunk.SystemFingerprint
 		}
 
-		// Keep the last chunk for metadata
-		lastChunkResponse = chunk
+		// Keep the last chunk with valid choices for metadata.
+		// Skip non-standard events (e.g. inference-cost) that have empty
+		// choices and would overwrite the real last chunk's ID/Model/Created.
+		if len(chunk.Choices) > 0 {
+			lastChunkResponse = chunk
+		}
 	}
 
 	// Create a complete ChatCompletionResponse based on the last chunk structure
--- a/llm/transformer/openai/outbound.go
+++ b/llm/transformer/openai/outbound.go
@@ -328,14 +328,16 @@ func (t *OutboundTransformer) TransformStreamChunk(
 		return nil, err
 	}
 
-	// Skip non-standard events with explicit empty choices array (e.g. inference-cost,
-	// cost) that some providers emit alongside standard chat completion chunks.
-	// Returning nil causes NoNil to filter it from the client stream.
-	//
-	// We only filter when the raw JSON has "choices":[] (present but empty).
-	// Events without a "choices" key (e.g. stub/aggregate-only events) are passed through.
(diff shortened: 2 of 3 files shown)


looplj:
感谢 PR，需要修一下 ci

Black0416:
已修复 CI 失败。

原来的 `len(resp.Choices) == 0` 过滤太宽泛了——`lo.Map(nil, ...)` 返回非 nil 空切片，导致 stub/aggregate 事件也被过滤。

改为用 `gjson` 检查原始 JSON：只在 `"choices":[]`（key 存在且为空数组）时才过滤，`"choices"` key 不存在时正常通过。

本地 `go test ./llm/...` 全部通过，包括之前失败的 `TestPipeline_NonStreaming_AutoAggregateUpgradedStream_EmptyJSONObjectAggregatedBodyAllowed`。

PR #2018, opened 2026-07-14
[codex] fix(apikeys): support clipboard fallback on HTTP

lihaojiee:
## Summary

- add a shared clipboard utility for API key copy actions
- use the modern Clipboard API when available and a DOM copy command when it is unavailable
- migrate API key table, view dialog, base URL, rotation result, and masked code example copy actions to the shared implementation
- show success feedback only after the text has actually been copied

## Problem

When AxonHub is opened from a non-secure HTTP origin, such as a LAN deployment, `navigator.clipboard` is unavailable. Clicking an API key copy button throws `Cannot read properties of undefined (reading 'writeText')`, so nothing is copied and the page provides no useful failure feedback.

The API key page also had several independent clipboard implementations with inconsistent success and error handling.

## Fix

- introduce `copyTextToClipboard` as the single low-level clipboard implementation
- prefer `navigator.clipboard.writeText` on secure origins
- use a temporary read-only textarea and `document.execCommand('copy')` only when the Clipboard API is unavailable
- throw when both copy mechanisms fail, allowing the shared hook to display the existing localized failure toast
- honor the hook's custom success message and update copied-state only after a successful copy

The legacy copy command is intentionally limited to non-secure origins where the modern Clipboard API is unavailable.

## Validation

- rebased the change onto the latest `upstream/unstable`
- IntelliJ inspections report no errors in the six changed files
- TypeScript `transpileModule` parsing passed for all changed TypeScript/TSX files
- `git diff --check` passed, including the new file
- confirmed that the repaired API key scope no longer calls `navigator.clipboard` directly

Full frontend lint/build was not run locally because the repository instructions prohibit running lint or build commands unless explicitly requested. The upstream PR CI will run the configured frontend build and repository checks.

### DIFF
--- a/frontend/src/components/ai-elements/masked-code-block.tsx
+++ b/frontend/src/components/ai-elements/masked-code-block.tsx
@@ -4,6 +4,7 @@ import { type ComponentProps, createContext, type HTMLAttributes, useContext, us
 import type { Element } from 'hast';
 import { CheckIcon, CopyIcon } from 'lucide-react';
 import { type BundledLanguage, codeToHtml, type ShikiTransformer } from 'shiki';
+import { copyTextToClipboard } from '@/lib/clipboard';
 import { cn } from '@/lib/utils';
 import { Button } from '@/components/ui/button';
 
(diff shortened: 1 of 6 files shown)


llc1123:
Closing as already implemented: #2334 (merged 2026-09-01) shipped the shared `frontend/src/lib/clipboard.ts` utility with the same `navigator.clipboard` → `document.execCommand('copy')` fallback for insecure HTTP origins, and migrated every call site this PR touched (`use-copy-to-clipboard.ts`, apikeys columns/rotate/view dialogs, masked-code-block). The merged version also adds stricter capability checks. Thanks for the contribution!

PR #2021, opened 2026-07-15
feat(channel): support insecure TLS verification

BugMasterLab:
## Summary
- add a per-channel option to skip HTTPS certificate verification while keeping secure verification as the default
- apply the option to both model discovery and saved-channel upstream requests without leaking TLS settings across channels
- preserve existing channel settings and global SSL behavior, with GraphQL, UI, locale, and focused test coverage

## Test plan
- [x] Verify secure clients still reject untrusted test certificates
- [x] Verify per-channel insecure clients accept untrusted certificates without mutating the shared client
- [x] Verify model fetching and channel requests honor the per-channel TLS setting
- [x] Build the frontend production assets with `pnpm -C frontend build`


looplj:
有冲突

PR #2027, opened 2026-07-17
fix: isolate upstream X-Request-ID per request

winterfell2021:
## Summary

- replace a client-provided `X-Request-Id` on outbound calls with AxonHub's unique per-request `ar-*` ID
- keep the ID stable for retry attempts belonging to the same inbound request
- keep channel header overrides authoritative by applying them after the generated request ID
- add regression coverage for multiple inbound requests that reuse one client request ID

## Problem

AxonHub trace IDs can intentionally span multiple agent calls, while an HTTP request ID must identify one request. The generic inbound-header merge currently forwards `X-Request-Id` unchanged. If a client reuses that value to group an agent trace, independent outbound calls receive the same request ID.

Downstream gateways that use `X-Request-Id` for usage idempotency or deduplication can then collapse unrelated requests into one record, mixing models, statuses, token usage, and charges.

AxonHub already generates a unique `ar-*` request ID for every inbound HTTP request in `WithLoggingTracing`; it was not previously applied to the outbound provider request.

## Fix

Add an outbound raw-request middleware that sets `X-Request-Id` from AxonHub's request context after inbound headers are merged. Different inbound requests therefore get different downstream IDs even when the client reuses its own value. Retries for one request keep the same ID, and explicit channel header overrides can still replace it.

## Verification

- `go test ./...`
- `cd llm && go test ./...`

### DIFF
--- a/internal/server/orchestrator/orchestrator.go
+++ b/internal/server/orchestrator/orchestrator.go
@@ -270,6 +270,9 @@ func (processor *ChatCompletionOrchestrator) Process(ctx context.Context, reques
 		// User-Agent value (either from client pass-through or default "axonhub/1.0").
 		// This allows override headers to modify the User-Agent if configured.
 		applyUserAgentPassThrough(outbound, processor.SystemService),
+		// Replace a client-provided X-Request-Id with AxonHub's unique per-request
+		// identifier. Header overrides run afterwards and can still opt out.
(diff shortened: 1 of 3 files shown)


looplj:
这个需要一个配置项
1. 是否开启
2. header 是什么

PR #2034, opened 2026-07-18
[codex] Forward verified reasoning accounting response header

asuan-dev:
## Summary

- Preserve upstream streaming handshake metadata through the HTTP client and pipeline.
- Forward `X-Reasoning-Included: true` to downstream clients for streaming, non-streaming, auto-aggregated, and Codex stream-to-JSON responses.
- Keep the boundary allowlisted: false/missing values and unrelated upstream headers are not forwarded.
- Ensure retries use only metadata from the final successful channel.

## Root cause

AxonHub terminates the upstream HTTP response, transforms the payload, and constructs a new downstream response. The streaming executor returned only `StreamEvent` values, so upstream response headers had no carrier. The API layer also rebuilt response headers rather than copying this protocol-significant header. The Codex non-stream adapter had a second stream-to-JSON aggregation boundary with the same loss.

This is not a full raw-header proxy by design: blindly copying hop-by-hop, credential-related, or provider-private headers would be unsafe. The fix adds a transport-level metadata carrier and an explicit safe forwarding rule.

## Validation

- `cd llm && go test ./httpclient ./pipeline ./transformer/openai/codex -count=1`
- `go test ./internal/server/orchestrator ./internal/server/api -count=1`
- `git diff --check`

The regression tests cover direct streaming, non-streaming, auto-aggregation, Codex aggregation, false/missing values, unrelated headers, and retry isolation.

### DIFF
--- a/internal/server/api/chat.go
+++ b/internal/server/api/chat.go
@@ -81,6 +81,8 @@ func (handlers *ChatCompletionHandlers) ChatCompletionWithRequest(c *gin.Context
 		return
 	}
 
+	writeForwardResponseHeaders(c, result.ResponseHeaders)
+
 	if result.ChatCompletion != nil {
 		resp := result.ChatCompletion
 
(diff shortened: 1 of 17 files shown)


asuan-dev:
按本地代码审查要求先关闭：将重新评估是否为最小改动、是否符合现有架构，并在完整验证通过后再决定是否重新提交。

PR #2035, opened 2026-07-18
fix(responses): forward verified reasoning accounting header

asuan-dev:
## Summary

- Preserve X-Reasoning-Included: true from successful upstream HTTP/SSE responses.
- Keep non-streaming metadata on the existing httpclient.Response.Headers.
- Carry streaming metadata through a small httpclient stream wrapper, without changing the generic streams.Stream interface or pipeline/orchestrator result structs.
- Preserve the header when the Codex executor aggregates a stream into a non-streaming response.
- Forward no header for missing, false, or conflicting values, and never forward unrelated headers such as Set-Cookie.

## Why

AxonHub terminates the upstream HTTP response and rebuilds the downstream response. The old path discarded the handshake header before it reached Codex, so Codex could re-estimate reasoning tokens that the server had already accounted for.

## Scope

This change covers the HTTP/SSE response path used by the current Codex channel. Successful WebSocket handshake response headers are not changed here; the WebSocket path has a separate metadata/event contract.

## Design

The allowlist and validation live in llm/httpclient. Non-streaming responses reuse the existing response header field. Streaming responses carry a cloned, filtered header alongside the stream and are re-wrapped after protocol transformations. The API boundary applies the same allowlist again as defense in depth.

## Validation

- cd llm && go test ./httpclient ./pipeline ./transformer/openai/codex -count=1
- go test ./internal/server/api ./internal/server/orchestrator -count=1
- git diff --check

### DIFF
--- a/internal/server/api/chat.go
+++ b/internal/server/api/chat.go
@@ -118,12 +118,15 @@ func (handlers *ChatCompletionHandlers) ChatCompletionWithRequest(c *gin.Context
 }
 
 func writeForwardResponseHeaders(c *gin.Context, result orchestrator.ChatCompletionResult) {
-	headers := httpclient.GetResponseHeaders(result.ChatCompletionStream)
+	var headers http.Header
 	if result.ChatCompletion != nil {
 		headers = result.ChatCompletion.Headers
+	} else {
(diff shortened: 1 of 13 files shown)


looplj:
需要帮忙确认下，本地是否验证过

PR #2039, opened 2026-07-20
fix(ent): align prompt project ownership

brynne8:
## Summary
- Bind prompts to the existing project_id ownership field and remove the unused M2M mapping.
- Default system roles to project ID 0 so the existing composite unique index protects direct Ent creation paths.
- Add Ent integration coverage for both invariants.

## Verification
- make generate
- go test ./internal/ent -run TestPromptProjectRelationship|TestSystemRoleUsesDefaultProjectID
- go test ./internal/server/biz -run TestCreateRole


looplj:
有点冲突

brynne8:
## Versioning follow-up

This PR targets the next release, v1.0.0-beta6.

- Source and local builds now fall back to v1.0.0-beta6 instead of the stale v0.9.38 file value.
- The legacy system-role migration is versioned v1.0.0-beta6, so databases recorded at v1.0.0-beta5 execute it during upgrade.
- Unstable Docker builds now use a SemVer-compatible v1.0.0-beta5-unstable.YYYYMMDD version. It remains ordered below beta6, so it cannot cause the beta6 migration to be skipped; this ordering is covered by a migration test.
- Tagged release artifacts continue to receive their exact tag through GoReleaser ldflags.

Longer term, data-migration completion should be tracked independently from the application version.

brynne8:
## Review follow-up

Duplicate legacy system roles now have an explicit migration policy:

- Roles with the same name and equivalent scopes are consolidated into a canonical role.
- User-role assignments are moved to that role, and redundant legacy rows are permanently removed.
- Roles with different scopes are rejected before any writes occur, preserving authorization semantics and leaving the transaction unchanged for manual resolution.

The foreign-key concern is intentionally outside the supported production migration configuration: startup creates schemas with WithForeignKeys(false), and the system-role project ID 0 sentinel relies on that established policy. Deployments that independently enforce the generated Role-to-Project FK are not compatible with system roles using this sentinel.

PR #2051, opened 2026-07-21
fix(responses): insert synthetic tool messages for missing function_call_output items

### LINKED ISSUE #2049: fix: Responses API inbound converter should handle missing function_call_output items
## Bug Description

When Codex (Responses API) sends conversation history that contains `function_call` items without matching `function_call_output` items, AxonHub proxies these directly to downstream providers (e.g., Kimi). The downstream provider rejects the request because the converted Chat Completions messages contain `tool_calls` without corresponding `tool` responses.

## Error Message

```
failed to stream request: Request failed: Bad Request, error: an assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: exec_command:10, type: invalid_request_error
```

## Root Cause

In `llm/transformer/openai/responses/inbound.go`, the `convertInputToMessages` function converts each `function_call` item to an independent assistant message with its `tool_calls`. It does not check whether a matching `function_call_output` exists for every `function_call`. When a `function_call` lacks its output, the resulting `llm.Messages` array violates Chat Completions API expectations.

Unlike the Responses API (which manages conversation state server-side), downstream Chat Completions providers (Kimi, etc.) strictly enforce tool call → tool response pairing and reject incomplete cycles.

## Proposed Fix

Add a post-processing step `fixMissingToolCallOutputs` that:
1. Collects all `tool_call_id`s that already have responses from `role: "tool"` messages
2. For each assistant message with `tool_calls`, inserts a synthetic `role: "tool"` message (with empty content) for any tool call that lacks a matching response

This is a defensive normalization that prevents malformed conversation history from causing downstream rejections.

ThinkPeace:
## Summary

When converting Responses API input items to Chat Completions messages, `function_call` items without matching `function_call_output` items produce assistant messages with `tool_calls` but no `tool` responses. Downstream providers reject such incomplete tool call cycles.

## Fix

Add `fixMissingToolCallOutputs` — a post-processing step in `convertInputToMessages` that:
1. Collects all `tool_call_id`s that already have responses
2. Inserts synthetic empty `role: "tool"` messages for any tool call lacking a response

## Changes

- **`llm/transformer/openai/responses/inbound.go`**: Add `fixMissingToolCallOutputs`, called at end of `convertInputToMessages`
- **`llm/transformer/openai/responses/inbound_test.go`**: Add 9 test cases

All existing tests in `transformer/openai/responses/` continue to pass.

Closes #2049

### DIFF
--- a/internal/server/orchestrator/orchestrator.go
+++ b/internal/server/orchestrator/orchestrator.go
@@ -75,6 +75,7 @@ func NewChatCompletionOrchestrator(
 		PromptProvider:     promptService,
 		PromptProtecter:    promptProtectionRuleService,
 		Middlewares: []pipeline.Middleware{
+			cc.FixMissingToolCalls(),
 			cc.StripBillingHeaderCCH(),
 			stream.EnsureUsage(),
 		},
--- a/llm/pipeline/cc/fix_tool_calls.go
+++ b/llm/pipeline/cc/fix_tool_calls.go
@@ -1,0 +1,67 @@
+package cc
+
+import (
+	"context"
+
+	"github.com/looplj/axonhub/llm"
+	"github.com/looplj/axonhub/llm/pipeline"
+	"github.com/looplj/axonhub/llm/transformer/openai/responses"
+)
+
+// FixMissingToolCalls returns a pipeline middleware that:
+//  1. Inserts synthetic tool messages for any assistant tool_calls that lack a
+//     corresponding tool response.
+//  2. Adds dummy tool definitions for any function name referenced in tool_calls
+//     that is missing from the request's tools array.
+//
+// This prevents downstream providers from rejecting requests due to incomplete
+// tool call cycles or unknown tool definitions (e.g., Codex-specific tools like
+// exec_command that are present in conversation history but not in the client's
+// tool definitions).
+func FixMissingToolCalls() pipeline.Middleware {
+	return pipeline.OnLlmRequest("FixMissingToolCalls",
+		func(ctx context.Context, request *llm.Request) (*llm.Request, error) {
+			if request == nil || len(request.Messages) == 0 {
+				return request, nil
+			}
+
+			// Step 1: Fix missing tool responses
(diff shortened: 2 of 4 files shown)


ThinkPeace:
Good catch! Fixed in 459675e.

The updated algorithm does **per-assistant-message independent checking** instead of using a global `responded` map:

1. For each assistant message, collect its `tool_call_id`s into a `missing` set
2. Look ahead at immediately-following `role: "tool"` messages (stop at first non-tool), removing covered IDs from `missing`
3. Insert synthetic tool messages for any remaining uncovered IDs

This means `[assistant(x), tool(x), assistant(x), user]` correctly gets a synthetic `tool(x)` inserted after the second assistant turn — the first turn's `tool(x)` is before the second assistant, not after it, so it does not cover the second call.

Added a dedicated test case: "same ID across two assistant turns — each checked independently" that validates this exact scenario.

ThinkPeace:
Updated with the full fix. The original approach only covered the `/v1/responses` path. Testing revealed the `/v1/chat/completions` path also needs the fix.

### Changes since last review

1. **Exported `FixMissingToolCallOutputs`** so it can be called from other packages
2. **Added `llm/pipeline/cc/fix_tool_calls.go`** — a pipeline middleware that runs on ALL request paths (both `/v1/responses` and `/v1/chat/completions`)
3. **Registered middleware in orchestrator** as a global middleware
4. **Added tool definition fix**: the middleware now also ensures every tool name referenced in conversation history has a matching entry in the `tools` array (prevents `tool call id X is not found` errors when Codex-specific tools like `exec_command` appear in history but not in the client's tool definitions)
5. **8 test cases** covering the exported function

### Middleware does two things on every request:

1. Calls `FixMissingToolCallOutputs` to insert synthetic tool messages for any assistant tool_calls lacking responses
2. Scans conversation history for tool names not in `request.Tools` and adds minimal dummy definitions

looplj:
这是客户端问题吧

PR #2067, opened 2026-07-23
fix(backup): stream backup JSON to prevent OOM on large datasets

djdembeck:
## Summary
Replace in-memory backup accumulation with streaming JSON writer to prevent OOM on large dataset backups

## Motivation
The previous `doBackup` method loaded entire entity tables into memory via `.All()` + `lo.Map` before serializing. With large usage logs or request histories, this could exhaust available memory and crash the process mid-backup — particularly problematic for auto-backup running unattended.

## Changes
- **Streaming JSON backup** (`backup_ops.go`): New `doBackupToWriter` method writes compact JSON directly to an `io.Writer`, processing each entity type in batch-500 ID-cursor pages instead of loading all rows at once
- **Auto-backup streaming path** (`autobackup.go`): Non-DB storage backends now stream directly to a temp file then upload via `SaveDataFromReader`, avoiding double-buffering in memory during backup
- **DB storage unchanged** (`autobackup.go`): Database-backed backup storage still uses the in-memory path since the DB driver manages its own buffers
- **Constant rename** (`restore.go`): `usageBackupBatchSize` → `backupBatchSize` to reflect broader use across backup operations
- **Tests**: Added pagination-across-batch-boundary test case

## Notes
- Empty optional fields in backup output now produce absent keys rather than `"field":[]` when included-but-empty, matching the `omitempty` struct tag behavior — restore code uses `len(x)==0` checks that treat nil and empty slices identically
- Removed `github.com/samber/lo` dependency from the backup package

### DIFF
--- a/internal/server/backup/autobackup.go
+++ b/internal/server/backup/autobackup.go
@@ -11,6 +11,7 @@ import (
 	"github.com/spf13/afero"
 
 	"github.com/looplj/axonhub/internal/ent"
+	"github.com/looplj/axonhub/internal/ent/datastorage"
 	"github.com/looplj/axonhub/internal/log"
 	"github.com/looplj/axonhub/internal/server/biz"
 	"github.com/looplj/axonhub/internal/server/scheduler"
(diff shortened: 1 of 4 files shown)


djdembeck:
Tested and working: memory usage stayed below 4GB when running the same backup that previously maxed 6GB of available RAM, plus 1GB of swap RAM. New system keeps axonhub from taking over RAM and stalling during backup now.

PR #2068, opened 2026-07-24
fix(orchestrator): detect stream completion from JSON data type and finish_reason

trim21:
Some SSE-compatible providers do not populate the SSE `event` field and instead carry the event type only in the JSON data payload. This also recognizes a chat completion's `finish_reason` as a semantic stream completion marker, since clients commonly close the connection immediately after consuming that final useful chunk, before the trailing `[DONE]` marker is read by the server.

## Motivation: real-world issue with pi (coding agent)

When using [pi](https://github.com/earendil-works/pi) (an AI coding agent) through AxonHub, streaming chat completion requests were incorrectly marked as `context canceled` errors even though the response was complete:

1. Pi uses the OpenAI SDK to stream chat completions through AxonHub.
2. The upstream provider returns a chunk with `finish_reason: "stop"` in the JSON data, but does **not** set the SSE `event` field and does not follow up with a `[DONE]` marker (or `[DONE]` arrives after pi has already processed `finish_reason`).
3. Pi's OpenAI SDK sees `finish_reason`, the `for-await` loop ends naturally, pi emits a `done` event, and **closes the SSE connection** to AxonHub.
4. At this point, AxonHub (the proxy) sees the client disconnect. Before this fix, `isTerminalStreamEvent` only checked the SSE `event` field and `[DONE]` byte comparison — it could not recognize `finish_reason` as a stream completion marker.
5. With `StreamCompleted = false` and `ctx.Done()`, AxonHub persisted the execution as failed/canceled.

After this fix, `isTerminalStreamEvent` recognizes `finish_reason` in the JSON data as a semantic completion, so `StreamCompleted` is set to `true` during `Next()`. When pi closes the connection, `Close()` sees `StreamCompleted = true` and correctly persists the execution as completed.

## Changes

- **inbound.go**: Parse JSON data for event types (`response.completed`, `message_stop`, `speech.audio.done`, `transcript.text.done`) and treat non-empty `finish_reason` in `choices` as stream completion. Add nil guard.
- **inbound_test.go**: Update existing test expectations and add `TestIsTerminalStreamEvent_SemanticCompletionInData` covering JSON-data-based detection.
- **outbound_test.go**: Add integration test (`canceled client after finish reason is still completed without done or usage`) verifying that a canceled context after `finish_reason` still completes the execution successfully.

### DIFF
--- a/internal/server/orchestrator/inbound.go
+++ b/internal/server/orchestrator/inbound.go
@@ -5,6 +5,8 @@ import (
 	"context"
 	"errors"
 
+	"github.com/tidwall/gjson"
+
 	"github.com/looplj/axonhub/internal/dumper"
 	"github.com/looplj/axonhub/internal/ent"
 	"github.com/looplj/axonhub/internal/log"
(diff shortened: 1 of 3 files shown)


looplj:
@greptile-apps review

PR #2120, opened 2026-07-31
feat: add per-channel API key rule actions

Lxiny-zy:
## Summary

- add per-channel API key rules matching HTTP status codes and response keywords/regular expressions
- support configurable consecutive-error thresholds, temporary disables with automatic recovery, and permanent disable/delete actions
- expose rule management and disable expiry through GraphQL and the channel UI
- preserve global auto-disable behavior as a fallback when no channel rule takes action

## Tests

- `go test ./internal/server/biz ./internal/server/orchestrator ./internal/server/gql`

## Summary by CodeRabbit

- **New Features**
  - Added configurable, channel-specific API key auto-disable rules based on status codes, keywords, and failure counts.
  - Supports temporary disables with automatic recovery or permanent disable/delete actions.
  - Manage rules directly from each channel’s settings.

- **Improvements**
  - Disabled API keys now show localized expiration information.
  - Expired temporary disables are cleaned up automatically, restoring channel availability when appropriate.

- **Documentation**
  - Added release notes describing API key rule behavior and recovery options.


Lxiny-zy:
@looplj    前端
image
image
测试了一下应该没bug？ 您可以看看

PR #2124, opened 2026-08-01
fix(transform): add channel switch to downgrade mid-conversation system messages

AkaChou:
## Summary

Fix prompt-cache cycle failures when Claude Code is routed through OpenAI-compatible upstreams (deepseek, zai/glm, moonshot, etc.).

Claude Code injects `role=system` reminder messages **mid-conversation** (task tool reminders, `UserPromptSubmit` hook context, plan-mode notices, system notifications). OpenAI-compatible upstreams hoist **all** system messages to the front of the prompt, so every newly injected reminder rewrites the whole system prefix and defeats prefix-based prompt caching. In production ~11% of requests drop to <10% cache-hit (normal ~90%).

## Root cause (verified against production data)

- Request bodies contain 10-19 `role=system` messages scattered through `messages` (same reminder repeated up to 5×), while the real system prompt lives in the top-level `system` field.
- 2×2 correlation: of 110 low-hit (<10%) requests, 108 ended with a `system` message; of 880 high-hit (>90%) requests, 877 did **not**.
- `llm.Request` has no separate `System` field — the inbound transformer places the top-level system at `Messages[0]`, indistinguishable from injected reminders downstream.

## Change

Add a per-channel switch `downgradeMidConversationSystem` (TransformOptions, **default off**) that downgrades mid-conversation `system` messages to `user` before sending, keeping the leading contiguous system block (the real system prompt) stable across turns.

- Consumed in `openai.RequestFromLLM` (the shared converter used by all OpenAI-compatible outbound transformers: deepseek, moonshot, gemini, openrouter, doubao, zai, openai, copilot) — one change covers all affected channels.
- Anthropic outbound is unaffected (does not go through `RequestFromLLM`); official Anthropic channels keep native `mid-conversation-system` support.
- Only the `role` label changes; content and position are preserved. No channel is affected by default.

## Files

- `llm/options.go`, `llm/transformer/openai/outbound_convert.go` (+test) — converter logic
- `internal/objects/channel.go`, `internal/server/orchestrator/transform_options.go` (+test) — channel config wiring
- `internal/server/gql/axonhub.graphql`, `internal/server/gql/generated.go` — GraphQL schema
- `frontend/src/features/channels/...` + locales — channel "Transform Options" dialog switch

## Testing

- Unit tests: `downgradeMidConversationSystem` (7 cases + multimedia content), `RequestFromLLM` switch tri-state (nil/false/true), `applyTransformOptions` tri-state.
- `go build ./...`, `go test` on openai/deepseek/moonshot/zai/orchestrator packages — all pass.
- golangci-lint (root + llm module): 0 issues.
- End-to-end (local): default-off keeps mid-system; explicit on downgrades `[system,user,system,user]` → `[system,user,user,user]` observed at upstream.

## Note on Transform Options interaction

`downgradeMidConversationSystem` and `replaceDeveloperRoleWithSystem` are **mutually exclusive in practice** - do not enable both on the same channel:

| Switch | Direction | Use when |
|---|---|---|
| `replaceDeveloperRoleWithSystem` | developer -> system | The upstream **does not support** the `developer` role (e.g. Bailian/Volc); promote developer instructions to system |
| `downgradeMidConversationSystem` | mid-conversation system -> user | Claude Code injects `system` reminders mid-conversation; OpenAI-compatible upstreams hoist all system messages to the front and break prompt caching |

If both are enabled, `applyTransformOptions` runs `replaceDeveloperRoleWithSystem` first, then `RequestFromLLM` runs `downgradeMidConversationSystem`, so a mid-conversation `developer` message becomes `system` and then `user` - the promotion is undone. They target opposite upstream behaviors (one promotes roles, the other demotes them), so a channel only ever needs one. Production channels currently enable `downgradeMidConversationSystem` alone (`replaceDeveloperRoleWithSystem = false`).

## Summary by CodeRabbit

* **New Features**
  * Added an option to convert system messages introduced mid-conversation into user messages.
  * The setting is available in channel configuration, persists across channel operations, and defaults to disabled.
  * Added English and Simplified Chinese labels and descriptions.

* **Bug Fixes**
  * Preserved leading system messages, other message content, and multimedia content during conversion.

* **Tests**
  * Added coverage for enabled, disabled, unset, empty, and multimedia message scenarios.


looplj:
有点冲突了， 需要帮忙更新下

AkaChou:
> 有点冲突了， 需要帮忙更新下

冲突已经处理～

PR #2137, opened 2026-08-03
fix(openai): keep message content valid for strict upstreams

Lxiny-zy:
## Problem

Two issues make requests fail against strict OpenAI-compatible upstreams when a Responses-protocol client (e.g. Codex CLI) is routed to a Chat Completions channel.

**1. Missing or null `content`**

An assistant message that only requests tool calls has no content, so `content` is omitted entirely by `omitzero`. A message whose content parts were all filtered out (compaction) serializes as `content: null` instead, because `lo.FilterMap` returns a non-nil empty slice and `MarshalJSON` falls through to the nil `Content` pointer.

Both are legal per the OpenAI spec, but upstreams whose schema types content as `string | ContentPart[]` reject them:

```
{"error":{"message":"body/messages/6/content Invalid input","type":"invalid_request_error"}}
```

This surfaces mid-conversation — the first turns succeed, and the request only fails once the assistant emits a turn with no text alongside its tool call.

**2. Responses-only content part types**

Responses distinguishes `input_text` from `output_text`; Chat Completions has a single `text` part type. `MessageContentPartFromLLM` copied the type verbatim, so upstreams saw part types they don't know.

Single text parts are collapsed into a plain string earlier in `convertToMessageContent`, which is why this only shows up with multi-part content (e.g. text alongside an image).

## Fix

- Normalize content to `""` for assistant messages that have tool calls but no usable content. OpenAI treats it the same as null, and strict upstreams accept it. The condition is scoped to messages with tool calls, so every other message keeps its current serialization.
- Map `input_text`/`output_text` onto `text`. Types Chat Completions does understand (`image_url`, `video_url`, `input_audio`) pass through untouched.

## Tests

Five new tests in `outbound_convert_test.go`:

- tool-call-only messages keep a `content` field, for both the missing-field and the all-parts-filtered cases
- existing content on a tool-call message is preserved
- messages without tool calls keep their current serialization (content stays omitted)
- part type normalization, including the types that must not be rewritten
- end-to-end through `RequestFromLLM` with multi-part content

`internal/server/orchestrator` and `internal/server/biz` suites pass. Three pre-existing failures in the `llm` module (two missing testdata fixtures, one Windows-specific websocket error string) are unrelated and reproduce on an unmodified tree.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with stricter Chat Completions providers when sending assistant messages containing tool calls.
  * Ensured tool-call messages use valid empty content when no text is present.
  * Preserved existing message content while filtering unsupported or empty content parts.
  * Normalized supported input and output text parts to the standard `text` format.
  * Preserved non-text content types and multipart message content where supported.

### DIFF
--- a/llm/transformer/openai/outbound_convert.go
+++ b/llm/transformer/openai/outbound_convert.go
@@ -186,6 +186,16 @@ func MessageFromLLMWithConfig(m llm.Message, reasoningField ReasoningField) Mess
 		})
 	}
 
+	// An assistant turn that only requests tool calls has no content to send, and
+	// a message whose parts were all filtered out (e.g. compaction) is left with an
+	// empty part list. Both cases would reach the wire as a missing or null content
+	// field, which the OpenAI spec permits but stricter OpenAI-compatible upstreams
+	// reject because their schema only accepts a string or an array. Normalize to an
+	// empty string, which every implementation accepts and OpenAI treats as no content.
+	if len(msg.ToolCalls) > 0 && msg.Content.Content == nil && len(msg.Content.MultipleContent) == 0 {
+		msg.Content = MessageContent{Content: lo.ToPtr("")}
+	}
+
 	// Convert Annotations
 	if len(m.Annotations) > 0 {
 		msg.Annotations = lo.Map(m.Annotations, func(a llm.Annotation, _ int) Annotation {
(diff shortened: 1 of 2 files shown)


Lxiny-zy:
@looplj 您可以看下，我使用的时候发现有的旧渠道的兼容存在一点这样的问题

PR #2166, opened 2026-08-06
fix(quota): color Codex quota bars by usage

djdembeck:
## Summary

Codex usage bars were colored from the reset-window elapsed time instead of the actual usage percentage, so low usage could appear red near the start of a window. The usage bars now color from the used percentage, while the separate duration bars keep showing reset-window elapsed time.

## Motivation

The severity color of the primary/secondary usage bars reflected how much of the reset window had elapsed, not how much quota had been consumed. Right after a window reset, usage bars could render in the red range even though usage was near zero, making them visually misleading.

## Changes

- Color the Codex primary and secondary usage bars from their used-percentage values.
- Keep reset-window elapsed time on the separate duration bars.
- Add a focused regression test asserting the Codex branch renders the used percentage into both usage bars, keeps two duration bars, and does not pass `durationPercentage` into the usage `ProgressBar`s.

## Notes

The dedicated regression test passes, and no matching upstream issue was found for this behavior during PR creation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Codex quota usage indicators to reflect actual primary and secondary usage levels.
  * Preserved separate reset-window duration indicators and reset information.
  * Corrected severity coloring so it is no longer influenced by reset-window progress.

### DIFF
--- a/frontend/src/components/quota-badges.test.mjs
+++ b/frontend/src/components/quota-badges.test.mjs
@@ -1,0 +1,57 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+const componentsDir = import.meta.dirname;
+const srcRoot = join(componentsDir, '..');
+
+function read(relativePath) {
+  return readFileSync(join(srcRoot, relativePath), 'utf8');
+}
+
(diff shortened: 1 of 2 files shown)


djdembeck:
@looplj if you can re-start the failed CI checks so we can see merge status that would be great :)

PR #2204, opened 2026-08-12
feat(quota): track OpenCode Go quota via official usage API

jveko:
## Summary

OpenCode now exposes an official usage endpoint, so quota tracking no longer needs to scrape the workspace dashboard with a `workspaceId` + auth cookie.

```
GET https://opencode.ai/zen/go/v1/usage
Authorization: Bearer <channel API key>
```

## Changes

**Backend**
- `provider_quota/opencode_go_checker.go`: replaced the dashboard HTML scrape (hydration `$R[n]` regex + `data-slot` DOM fallback + human-duration parser) with a JSON call to `/zen/go/v1/usage` using the channel's existing API key. `resetsAt` accepts RFC3339 with fractional seconds (live format), unix seconds, or epoch milliseconds. Raw `windows` shape (`usage_percent`, `reset_in_seconds`, `reset_time`, `status`, `percent_remaining`) is unchanged, so quota badges and limits keep working as-is; the API's per-window `status` is surfaced as `api_status` for observability.
- `provider_quota.go`: removed the auth-cookie credential gate; OpenCode Go channels now fall through to the standard API-key credential check.
- Removed the now-obsolete `settings.providerQuota.opencodeGo` (`workspaceId`/`authCookie`) config: `objects.ChannelSettings`, GraphQL schema, gqlgen mappings, regenerated `generated.go`.

**Data migration**
- New versioned data migrator `V1_0_0_Beta9` (`internal/ent/migrate/datamigrate/v1.0.0-beta9.go`) purges the stale `settings.providerQuota` key from existing channels on upgrade — dialect-aware (`#- '{providerQuota}'` on Postgres, `json_remove` on SQLite), idempotent, only touches rows that have the key. This removes the leftover auth cookie credential from the DB instead of leaving it until the next channel save. Verified against the maxim production DB: 2 rows currently carry the key.

**Frontend**
- Removed the Workspace ID / Auth Cookie fields from the channel dialog, the `providerQuota` query fragments, zod schemas, merge util, and i18n keys.
- Quota popover: dropped the per-workspace row collapsing (`openCodeGoWorkspaceKey`) — quota is now per API key, so each channel renders its own row. The rolling/weekly/monthly bars are unchanged.

## Verification

- **Live API** (verified 2026-08-12 against a real channel key): `GET /zen/go/v1/usage` returns `200` with
  ```json
  {"usage":{"rolling":{"status":"ok","percent":0,"resetsAt":"2026-08-12T11:24:29.905Z"},"weekly":{"status":"ok","percent":21,"resetsAt":"2026-08-17T00:00:00.905Z"},"monthly":{"status":"ok","percent":38,"resetsAt":"2026-08-31T23:33:35.905Z"}}}
  ```
  Note the millisecond-precision `resetsAt` — the checker parses it with `RFC3339Nano`. An invalid key returns `401 {"type":"error","error":{"type":"AuthError","message":"Unauthorized"}}`, which the httpclient surfaces as a request error.
- **Review pass** (2026-08-12, two independent reviewer agents over the full diff): no release-blocking findings. Hardened per review: epoch-milliseconds `resetsAt` (previously would silently parse as a ~year-58444 timestamp), API `status` surfaced as `api_status`, and closed the parser/401/edge test gaps.
- `go build ./...` — clean
- `go test ./internal/server/biz/provider_quota/...` — clean (17 tests: success incl. millis + min NextResetAt, unix-seconds and epoch-ms `resetsAt`, percent 100/120 exhaustion, partial windows, all-windows-unparseable, 401, missing key, multi-key fallback, `api_status` surfacing, `parseOpenCodeGoResetsAt` unit table, `SupportsChannel`)
- `go test ./internal/ent/migrate/datamigrate/...` — clean (3 new tests: strips `providerQuota` while preserving other settings, idempotent, untouched rows unaffected)
- `tsc --noEmit` (frontend, pinned TS 5.8.3) — clean

No schema migration needed (settings is a JSON column; ent schema untouched — per repo rule, no manual schema-migration SQL).

## Summary by CodeRabbit

- **New Features**
  - OpenCode Go usage quotas are now retrieved through the provider’s usage API using channel API keys.
  - Quota results include rolling, weekly, and monthly usage windows, reset times, and API status details.
  - Reset times are supported in multiple timestamp formats for more reliable quota displays.

- **Bug Fixes**
  - Improved handling of expired, partial, invalid, unauthorized, and over-limit quota responses.
  - Removed obsolete workspace and authentication-cookie configuration while preserving existing channel settings during migration.


jveko:
## CodeRabbit review — triage and fixes

**`CheckQuota` ignores HTTP status code (Major)** — **refuted, no change.** `httpclient.Do` (llm/httpclient/client.go:241-272) returns a typed `*Error{StatusCode, Status, Body}` for any `rawResp.StatusCode >= 400`, so `hc.Do` errors before `parseResponse` ever runs; the wrapped error (`opencode go usage request failed: ...`) carries the status code, and `parseResponse` only ever sees 2xx bodies. `TestOpenCodeGo_CheckQuota_Unauthorized` returns HTTP 401 from the transport and asserts the wrapped status error — it passes. (3xx are followed by the HTTP client, so the success path is effectively 2xx-only.)

**Reuse `ChannelCredentials.GetAllAPIKeys` (nit)** — **fixed.** `opencodeGoAPIKey` now iterates `GetAllAPIKeys()`, which applies the legacy-key ordering and the OAuth exclusion (an OAuth JSON in `apiKey` is no longer sent as a Bearer token), keeping only trim-and-skip-blank here.

**SQLite `json_type` predicate (Minor)** — **fixed.** The migration now uses `json_type(settings, '$.providerQuota') IS NOT NULL` so a present JSON-null `"providerQuota": null` is purged too (json_extract maps JSON null to SQL NULL, missing it). Added `TestV1_0_0_Beta9_StripsJsonNullProviderQuota`.

**Docstring coverage check** — **fixed.** Doc comments added for the exported symbols in `opencode_go_checker.go` (`OpenCodeGoUsageWindow`, `OpenCodeGoQuotaChecker`, `NewOpenCodeGoQuotaChecker`, `CheckQuota`, `SupportsChannel`).

Validation: `go test ./internal/server/biz/provider_quota/... ./internal/ent/migrate/datamigrate/...` clean, `go build ./internal/...` clean.

jveko:
## Full re-review (current HEAD, incl. data migration) — triage

**High — migration `*entsql.Driver` assert panics under read-replica config** — **fixed.** Confirmed real: `internal/server/db/ent.go` wraps the ent client driver in a `routerDriver` when `db.read_replica.read_dsn` is set, so `client.Driver().(*entsql.Driver)` fails and the migration error propagates to a startup panic. The migration now uses the `dialect.Driver.Exec` interface (satisfied by both `*entsql.Driver` and `routerDriver`, which routes writes to master), with `args []any{}` + `*sql.Result`. Tests green.

**Medium — Postgres jsonb path untested** — **verified live.** Ran the exact Postgres statement against the maxim production DB inside a rolled-back transaction: `UPDATE channels SET settings = settings #- '{providerQuota}' WHERE settings ? 'providerQuota'` → `UPDATE 2` (matches the 2 stale rows), `ROLLBACK`. Operators confirmed valid on the jsonb column.

**Medium — migration re-runs every boot until VERSION is bumped** — **accepted, documented.** `build.VERSION` is `v1.0.0-beta8`; this is a `v1.0.0-beta9` migration, so it runs idempotently on each boot until the release bumps VERSION to beta9 (VERSION is bumped via separate release commits, per git history). Now panic-free, so re-runs are a harmless no-op (0 rows after the first).

**Medium — MySQL silently skips the purge** — **accepted, documented.** MySQL is a supported dialect but not exercised here; the SQLite `json_type(...)` two-arg form is not valid MySQL (MySQL `JSON_TYPE` takes one arg), so I deliberately did not ship untested SQL that could panic MySQL boots at migration time. The default branch logs and skips. If MySQL is a real target, a follow-up can add a tested `JSON_TYPE(JSON_EXTRACT(...))` branch.

**Medium — OAuth/auth-json opencode channels poll-error each cycle** — **accepted.** opencode_go channels carry API keys, not OAuth; an OAuth-stamped opencode channel now falls through the generic credential gate and fails cleanly with `channel has no API key` + error backoff, identical to every other keyless provider. No prod impact (channel 34 has a real key in `apiKeys`).

**Low — parser overflow/boundaries** — **fixed.** `parseOpenCodeGoResetsAt` now rejects `asSeconds` outside `[0, 1e15]` (removing the unreachable `isFinite` helper and the `math` import), and tests added for the `1e12` seconds↔ms boundary (`999999999999` / `1000000000000`), `1e300`, negative percent, and the `GetAllAPIKeys` OAuth exclusion.

Validation: `go test ./internal/server/biz/provider_quota/... ./internal/ent/migrate/datamigrate/...` clean, `go build ./internal/...` clean, Postgres statement verified against prod (rolled back).

PR #2232, opened 2026-08-14
feat: dashboard statistics no longer lost when usage logs are cleaned up

Likhixang:
## Summary

Dashboard and analytics statistics are currently computed on-the-fly from the `usage_logs` detail table. The built-in GC cleanup (storage policy) deletes `usage_logs` rows older than the retention window, which **permanently erases historical statistics** from the dashboard — request counts, token usage, cost charts, top models/channels/API keys, and the analytics page all shrink or go empty after every cleanup run. There is no way to keep disk usage bounded without losing statistics.

This PR decouples statistics from log retention by adding a **day-granularity aggregate table (`usage_stats`)** that the dashboard reads instead of the detail logs. GC can keep purging old `usage_logs` rows freely; statistics survive indefinitely.

## Problem (Current State)

- All 16 dashboard/analytics aggregation queries (requests, tokens, cost, daily stats, top-N rankings, per-user usage) run `COUNT`/`SUM ... GROUP BY` directly against `usage_logs`.
- The GC worker (`internal/server/gc`) deletes `usage_logs` older than `storage_policy.usage_logs.cleanup_days` — either on a cron schedule or manually via the UI "clean now" button.
- Result: after cleanup, the dashboard permanently loses all history older than the retention window. The data is gone and cannot be recovered (a common complaint path is issue #2216, which shares the same root area: GC cleanup interacting badly with usage data).

## Solution

1. **New `usage_stats` table** — one row per `(local_date, api_key_id, model_id, channel_id, project_id)` holding aggregated `request_count`, token sums, and cost. Unique index on the natural key; 0 is used for absent `api_key_id`/`channel_id` so the unique index works on all supported databases (NULLs never conflict).
2. **Real-time upsert on write** — `UsageLogService.CreateUsageLog` now atomically increments the matching aggregate row (`ON CONFLICT ... DO UPDATE` with additive `+`). Failures are logged and never fail the request; the detail row remains the source of truth.
3. **One-time backfill on startup** — before the GC scheduler registers, `BackfillUsageStats()` aggregates the existing `usage_logs` into `usage_stats` (batched, idempotent, skipped when aggregates already exist). Historical statistics are preserved from the moment of upgrade onward.
4. **Dashboard/analytics read the aggregate** — all dashboard and analytics resolvers (`dashboard.resolvers.go`, `dashboard_helpers.go`, `analytics_helpers.go`, `analytics.resolvers.go`) now query `usage_stats` with date-range filters instead of `usage_logs` timestamp filters. GraphQL response shapes are unchanged, so the frontend needs no changes.
5. **GC unchanged** — `cleanupUsageLogs` still deletes detail rows only; it cannot touch statistics anymore.

## Key Design Points

- **Time window semantics**: day/week/month filters compare against the local calendar date string (`YYYY-MM-DD`) that the aggregate row is keyed on. Whole-day windows behave identically to the previous `created_at >= since` semantics. Arbitrary sub-day ranges (custom start/end times) are rounded to the enclosing day.
- **Scope of the change**: only statistics queries were migrated. Detail views (request log list/detail, thread/trace usage metadata) and quota enforcement (short minute/hour windows, unaffected by day-scale retention) still read `usage_logs`.
- **Time zone**: the aggregate date is computed with the system-configured `TimeLocation`, matching how the dashboard already derives its calendar periods.
- **GraphQL**: the new entity is exposed as a read-only Node (matching the auto-generated pattern of other ent schemas) with the same query policy as `usage_logs`; no frontend query uses it.

## Testing

- `go build ./...` — clean
- `go vet ./internal/server/...` — clean
- New integration tests (`internal/server/biz/usage_stat_test.go`, SQLite via `enttest`):
  - `TestUsageStat_UpsertAccumulates` — repeated upserts fold into a single row with additive counters
  - `TestUsageStat_Backfill` — aggregation by date/dimension is correct and idempotent
  - `TestUsageStat_SurvivesLogCleanup` — **deleting all `usage_logs` rows leaves the aggregates fully queryable** (the core guarantee of this PR)
- Existing `internal/server/gql` tests — pass

## Known Behavior Changes

- History older than the current retention window becomes visible again **after** backfill, and stays visible forever after that. Data already deleted by GC before this upgrade cannot be recovered (no source exists).
- Custom analytics ranges with sub-day granularity are clamped to whole days.
- `usage_stats` grows with the number of distinct (date × dimension) combinations, not with request volume — roughly `requests_per_day × active dimensions` rows, which is orders of magnitude smaller than `usage_logs`.

## Related

- #2216 — GC cleanup vs. usage data integrity issue (different root cause, but this PR makes the dashboard immune to the class of problems where cleanup corrupts or erases usage statistics; the aggregate table has a stable natural key and is not affected by request-ID reuse).

## Summary by CodeRabbit

* **New Features**
  * Added daily usage statistics aggregated by API key, project, channel, and model.
  * Added GraphQL querying, filtering, sorting, and pagination for usage statistics.
  * Added automatic aggregation for new usage logs and backfill for existing records.
* **Improvements**
  * Updated analytics and dashboard reporting to use aggregated usage data.
* **Bug Fixes**
  * Usage statistics remain available after usage-log cleanup.
  * Aggregation and backfill failures no longer interrupt requests or server startup.


Likhixang:
Closing this PR. PR #2246 has been merged with a better approach to the same problem — GC cleanup now strips stored request/response payloads rather than deleting rows, which preserves statistics without needing a separate aggregate table. The approach in #2246 is cleaner, more general, and already merged. Thanks!

PR #2251, opened 2026-08-17
fix: forward image parameter in /v1/images/generations

clhey:
## Problem

The `/v1/images/generations` endpoint silently dropped the `image` parameter when forwarding to upstream channels. This affected `gpt-image-1` image-to-image generation where input images are sent as data URLs in the JSON request body.

## Root Cause

Two issues in the OpenAI image transformer:

1. **Inbound** (`image_inbound.go`): `ImageGenerationRequest` struct had no `Image` field, so the `image` parameter in the JSON body was silently discarded during parsing.

2. **Outbound** (`image_outbound.go`): `buildImageGenerateRequest` only handled text-generation parameters (prompt, size, quality, etc.) and completely ignored `chatReq.Image.Images`, not forwarding input images to the upstream channel.

### Inbound
- Added `Image json.RawMessage` field to `ImageGenerationRequest`
- Added `parseGenerationImageField` to parse the `image` field (supports a single data URL string or an array of data URLs)
- Added `decodeDataURLToBytes` helper to decode data URLs to raw bytes
- Parsed images are stored in `llm.ImageRequest.Images`

### Outbound
- `buildImageGenerateRequest` now encodes `img.Images` back to data URLs and includes them in the request body as the `image` field
- Single image → string, multiple images → array (matching OpenAI API format)
- Added `encodeImageBytesToDataURL` helper

## Tests

Added 8 test cases:
- Inbound: single image, multiple images, invalid type, non-data-URL, round-trip
- Outbound: single image, multiple images, no image (backward compatibility)

All existing tests continue to pass.

## Summary by CodeRabbit

* **New Features**
  * Image generation requests can now include one or multiple input images.
  * Input images are accepted as base64 data URLs and forwarded in the request.
  * Generated image requests preserve image data using a single value or an array, as applicable.
* **Bug Fixes**
  * Invalid image formats, data URLs, and base64 content now return clear invalid-request errors.
* **Tests**
  * Added coverage for single-image, multiple-image, missing-image, invalid-input, and round-trip scenarios.

### DIFF
--- a/llm/transformer/openai/image_inbound.go
+++ b/llm/transformer/openai/image_inbound.go
@@ -51,21 +51,24 @@ var allowedImageTypes = []string{
 }
 
 // ImageGenerationRequest represents the request structure for image generation API.
+// The Image field accepts a data URL string or an array of data URL strings for
+// image-to-image generation (supported by gpt-image-1).
 type ImageGenerationRequest struct {
-	Prompt            string `json:"prompt"`
-	Model             string `json:"model"`
-	N                 *int64 `json:"n,omitempty"`
-	Quality           string `json:"quality,omitempty"`
-	ResponseFormat    string `json:"response_format,omitempty"`
-	Size              string `json:"size,omitempty"`
-	Style             string `json:"style,omitempty"`
-	User              string `json:"user,omitempty"`
-	Background        string `json:"background,omitempty"`
-	OutputFormat      string `json:"output_format,omitempty"`
-	OutputCompression *int64 `json:"output_compression,omitempty"`
-	Moderation        string `json:"moderation,omitempty"`
-	PartialImages     *int64 `json:"partial_images,omitempty"`
-	Stream            bool   `json:"stream,omitempty"`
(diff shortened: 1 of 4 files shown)


clhey:
image

image

clhey:
@looplj @sxueck please review, this pr fix /v1/images/generations endpoint silently dropped the image parameter bug

llm/transformer/openai/image_inbound.go
```diff
@@ -546,3 +555,58 @@ func parseOptionalInt64(s string) *int64 {
...
+		}
+
+		images = append(images, data)
+	}
+
+	return images, nil
+}
+
+// decodeDataURLToBytes decodes a data URL (data:image/png;base64,...) to raw bytes.
+func decodeDataURLToBytes(dataURL string) ([]byte, error) {
```
looplj, line 596: 看下是否可以复用 xurl

clhey: 可以的，我重新修改了，帮忙再看看

PR #2266, opened 2026-08-18
fix: skip postgres monotonic updated_at cleanup

djdembeck:
## Summary
AxonHub can fail to start on PostgreSQL after the `v1.0.0-beta9` migration because the monotonic `updated_at` cleanup was written for SQLite and then run against a typed `timestamptz` column. This PR limits that cleanup to SQLite and adds regression coverage for the PostgreSQL path.

## Motivation
The migration introduced by commit `86eac3a7` began applying a string-based `LIKE '% m=%'` rewrite to `channels.updated_at` for every supported dialect. On PostgreSQL, that column is `timestamptz`, so the statement fails with `operator does not exist: timestamp with time zone ~~ unknown` and the server panics before startup completes.

## Changes
- The `v1.0.0-beta9` migration now only rewrites `channels.updated_at` for SQLite.
- PostgreSQL keeps the original `settings.providerQuota` purge and skips the `updated_at` cleanup with a logged notice.
- Added a regression test proving the PostgreSQL migration executes only the JSONB purge and does not attempt the string-based `updated_at` cleanup.

## Notes
- This is the minimal fix for the production startup failure.
- The breaking commit is `86eac3a7` ("fix(channel): write UTC updated_at when saving model prices (#2257)").
- The new test exercises the migration through a fake driver rather than a live PostgreSQL instance.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database migration behavior for PostgreSQL by skipping an unnecessary timestamp cleanup.
  * Preserved timestamp cleanup for SQLite and existing behavior for unsupported database types.

### DIFF
--- a/internal/ent/migrate/datamigrate/v1.0.0-beta9.go
+++ b/internal/ent/migrate/datamigrate/v1.0.0-beta9.go
@@ -86,13 +86,16 @@ func (v *V1_0_0_Beta9) purgeProviderQuota(ctx context.Context, client *ent.Clien
 func (v *V1_0_0_Beta9) stripMonotonicUpdatedAt(ctx context.Context, client *ent.Client, dialectName string) error {
 	var stmt string
 	switch dialectName {
-	case dialect.Postgres:
-		stmt = `UPDATE channels SET updated_at = split_part(updated_at, ' m=', 1) WHERE updated_at LIKE '% m=%'`
 	case dialect.SQLite:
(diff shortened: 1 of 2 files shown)


brynne8:
@djdembeck @looplj 谢谢修复，不过我不懂了，为什么会有beta9的迁移脚本？明明官方版本号才到beta7。难道不应该所有迁移都在beta8吗？是不是vibe coding的问题？我之前让AI写代码的时候也遇到过它瞎提升版本号的事情。被我发现，所以才纠正了，大概是beta6的时候。

PR #2243, opened 2026-08-16
fix(codex): accept completed JSON Responses from compatible relays

MrLiuGangQiang:
## Summary / 摘要

This fixes a protocol-compatibility failure in the Codex outbound executor when a downstream client requests a completed Responses result (`stream: false`) but a Codex-compatible relay returns a normal completed JSON document instead of SSE.

Previously, the Codex outbound transformer correctly converted the upstream request to `stream: true` and negotiated `Accept: text/event-stream`, but the non-stream executor always consumed the upstream response through `DoStream()`. When a compatible relay returned `application/json` (including `application/json; charset=utf-8`), the SSE decoder produced zero events and aggregation failed with:

```text
failed to do request: empty stream chunks
```

## Root cause / 根因

The failure is not an empty model response. It is a transport mismatch:

1. The client sends a non-stream Responses request.
2. The Codex transformer still enables upstream streaming, as required by the Codex path.
3. Some compatible proxies/relays return a completed Responses JSON object instead of an SSE event sequence.
4. The old non-stream executor expected SSE unconditionally, collected no chunks, and surfaced `empty stream chunks`.

## What changed / 修改内容

- Execute a non-compact non-stream Codex request **once** through `inner.Do()` and inspect the completed response.
- If the upstream media type is `application/json` or an `application/*+json` type, return the response body unchanged so the existing Responses response transformer handles it.
- For every non-JSON response, preserve the previous behavior: decode the already-read body as SSE, retain top-level WebSocket error handling, and aggregate the events into a completed JSON response.
- Keep the compact-request path and the client-facing streaming (`DoStream`) path unchanged.

## Duplicate-execution safety / 避免重复执行

The implementation intentionally does **not** retry the request with another protocol after receiving a response. Retrying could run the model twice and create duplicate usage/billing. The decision is made from the single received response's `Content-Type`.

## Regression coverage / 回归测试

Added focused HTTP-server regression coverage that verifies all of the following:

- A downstream request with `stream: false` is still normalized to an upstream request with `stream: true` and `Accept: text/event-stream`.
- A completed Responses object returned as both `application/json; charset=utf-8` and `application/vnd.compat+json; charset=utf-8` is returned successfully instead of producing `empty stream chunks`.
- The completed JSON response continues through the existing Responses transformer and yields the expected `OK` assistant output.
- Exactly one upstream request is sent for each response type.
- The test SSE serializer emits a separate `data:` line for every payload line, including CRLF input, so existing SSE aggregation tests remain protocol-faithful.

[...]

PR #2278, opened 2026-08-20
fix(codex): scope Responses Lite to explicit official clients

MrLiuGangQiang:
﻿## Summary

Fix Codex Responses handling for ordinary OpenAI-compatible clients and compatible relays.

- keep official Codex SSE aggregation while allowing compatible relays to return one completed Responses JSON document for non-stream callers;
- do not fabricate `X-OpenAI-Internal-Codex-Responses-Lite` for ordinary clients;
- do not inject `reasoning.context: all_turns` into client requests;
- preserve explicitly selected Responses Lite only for the official `chatgpt.com` Codex backend;
- exclude Responses Lite from pass-through header replay so it cannot override the transformer’s target-aware decision.

## Why

Responses Lite is a private protocol mode with model- and tool-specific constraints. Treating it as mandatory identity metadata caused ordinary requests to fail with unsupported `reasoning.context` or unsupported tool errors.

## Tests

- Added coverage for ordinary requests, explicit official Lite requests using `current_turn`, and relay requests.
- Ran `gofmt` and `git diff --check`.
- Local Go tests could not start because this environment cannot reach `proxy.golang.org` to download Go modules; GitHub Actions will run the full suite.

## Summary by CodeRabbit

* **New Features**
  * Improved Codex compatibility across official backends and compatible relays.
  * Added support for handling both JSON and server-sent event responses from relay connections.
  * Preserved Responses Lite behavior for official backends while preventing unsupported forwarding to relays.

* **Bug Fixes**
  * Prevented internal protocol-selection headers from being forwarded incorrectly.
  * Removed automatic reasoning-context injection where it could conflict with request settings.

### DIFF
--- a/internal/server/orchestrator/pass_through.go
+++ b/internal/server/orchestrator/pass_through.go
@@ -17,9 +17,9 @@ import (
 	"github.com/looplj/axonhub/llm/transformer"
 )
 
-// codexResponsesPassThroughHeaders contains client metadata that Codex-compatible
-// Responses upstreams use to select protocol behavior. Keep this as an explicit
-// allowlist: inbound credentials and transport headers must never be copied.
+// codexResponsesPassThroughHeaders contains Codex identity metadata that can
+// accompany a pass-through body. Keep this as an explicit allowlist: inbound
+// credentials, transport headers, and protocol-selection headers are never copied.
 var codexResponsesPassThroughHeaders = []string{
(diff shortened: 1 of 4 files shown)


MrLiuGangQiang:
Superseded by the cleaned-up update to PR #2243; no duplicate PR needed.

PR #2294, opened 2026-08-22
fix(openai): synthesize terminal chunk for complete streams

Moyucharm:
## Summary

Some OpenAI-compatible upstreams emit the complete response content and a
usage chunk, then close the SSE stream without sending `finish_reason` or
`[DONE]`. Strict OpenAI-compatible clients interpret this as an incomplete
stream.

This affects coding-agent clients such as pi when the upstream response is
otherwise complete.

## Changes

- Add a stateful finalization layer to the OpenAI Chat Completions inbound
  transformer.
- When the source stream ends cleanly after producing output and usage, synthesize
  a final chunk with an empty `delta` and:
  - `finish_reason: "stop"` for text/reasoning output
  - `finish_reason: "tool_calls"` for tool-call output
- Emit `[DONE]` after a synthesized terminal chunk.
- Preserve existing `finish_reason` and `[DONE]` behavior.
- Do not synthesize success when the source has an error, usage is absent, or no
  meaningful output was produced, so genuinely truncated streams remain
  distinguishable.

## Reproduction

The affected upstream shape is:

1. One or more normal content/tool-call chunks.
2. A `choices: []` usage chunk.
3. Clean SSE EOF without `finish_reason` or `[DONE]`.

## Testing

Added coverage for:

- text streams missing `finish_reason`
- tool-call streams missing `finish_reason`
- reasoning-only output
- existing finish reasons
- existing `[DONE]`
- upstream errors
- missing usage
- partially finished multi-choice streams

Validation:

- `cd llm && go test ./transformer/openai/...`
- `cd llm && go vet ./transformer/openai/...`
- `go test ./internal/server/api/... ./internal/server/orchestrator/...`
- `go vet ./internal/server/api/... ./internal/server/orchestrator/...`

Follow-up to #1924; this complements #2185 by handling complete weak-provider
streams rather than merely reporting incomplete streams.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenAI streaming reliability by ensuring completed responses include the appropriate finish status.
  * Correctly identifies whether responses finish with normal text output or tool calls.
  * Preserves existing finish information and prevents duplicate completion events.
  * Prevents incomplete or errored streams from being incorrectly marked as complete.
* **Tests**
  * Added coverage for text, tool-call, reasoning-only, incomplete, and error scenarios.

### DIFF
--- a/llm/transformer/openai/inbound.go
+++ b/llm/transformer/openai/inbound.go
@@ -104,6 +104,8 @@ func (t *InboundTransformer) TransformStream(
 	ctx context.Context,
 	stream streams.Stream[*llm.Response],
 ) (streams.Stream[*httpclient.StreamEvent], error) {
+	stream = newOpenAIInboundStream(stream)
+
 	return streams.NoNil(streams.MapErr(stream, func(chunk *llm.Response) (*httpclient.StreamEvent, error) {
 		return t.TransformStreamChunk(ctx, chunk)
 	})), nil
--- a/llm/transformer/openai/inbound_stream.go
+++ b/llm/transformer/openai/inbound_stream.go
@@ -1,0 +1,233 @@
+package openai
+
+import (
+	"sort"
+	"strings"
+
+	"github.com/looplj/axonhub/llm"
+	"github.com/looplj/axonhub/llm/streams"
+)
+
+// openAIInboundStream finalizes complete OpenAI Chat Completions responses when
+// an upstream provider closes a clean stream without finish_reason or [DONE].
+// Usage is used as the positive completion signal so truncated streams are not
+// silently converted into successful responses.
+type openAIInboundStream struct {
+	source     streams.Stream[*llm.Response]
+	queue      []*llm.Response
+	queueIndex int
+	current    *llm.Response
+	finalized  bool
+
+	responseID    string
+	responseModel string
+	responseTime  int64
+	usage         *llm.Usage
+	doneSeen      bool
+	choices       map[int]*openAIInboundChoice
+}
+
+type openAIInboundChoice struct {
+	index        int
+	hasOutput    bool
+	hasToolCalls bool
+	finished     bool
+}
+
+func newOpenAIInboundStream(source streams.Stream[*llm.Response]) streams.Stream[*llm.Response] {
+	return &openAIInboundStream{
+		source:  source,
+		choices: make(map[int]*openAIInboundChoice),
+	}
+}
+
+func (s *openAIInboundStream) Next() bool {
+	if s.nextQueued() {
+		return true
+	}
+
+	s.queue = nil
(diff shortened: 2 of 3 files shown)


cjjdaq:
Hi, thanks for working on synthesizing terminal chunks for incomplete streams. This directly relates to a problem I'm hitting on the **OpenAI Responses + pass-through + streaming** path:

With `passThroughBody` enabled on a channel, an OpenAI Responses streaming request ends with a `STREAM_ERROR` event ("stream ended without terminal event or completed response") even though the upstream (e.g. tokenrhythm/DeepSeek) sends a complete, well-formed `response.completed`. The SSE writer's `terminalSeen` stays false because the raw provider stream is fanned out through `captureRawProviderStream` into the raw channel, and the writer never observes the terminal event — then `writeSSEStreamEnd` (chat.go) injects a synthetic error event to the client.

Would this PR's synthesize-terminal-chunk idea also cover the **Responses protocol + pass-through streaming** path (not just Chat Completions)? If not, is there a preferred way to handle the terminal event in the raw pass-through stream (e.g. skip the `terminalSeen` failure injection when pass-through is applied)?

Moyucharm:
Thanks for the detailed report — this PR would not cover that path.

The change here is scoped to the OpenAI Chat Completions inbound transformer. It synthesizes a missing finish_reason and [DONE] after a transformed Chat Completions stream reaches a clean EOF with positive completion evidence. With passThroughBody enabled, applyPassThroughStream replaces the transformed stream with the raw provider stream, so the synthesized chunks from this PR are intentionally bypassed.

Your case also sounds different semantically: the upstream already emitted a valid response.completed, so nothing should be synthesized. The raw fan-out should preserve that event, and the SSE writer should recognize it through IsTerminalStreamEvent, either from the SSE event: response.completed field or from type: "response.completed" in the JSON data.

I would avoid skipping the terminalSeen check for all pass-through streams, since that would also make genuinely truncated Responses streams appear successful and would weaken the incomplete-stream protection added by #2185.

The preferred fix seems to be on the raw pass-through terminal-observation path:

verify that captureRawProviderStream delivers the response.completed event to the raw channel;

verify that the event retains its Type and Data when it reaches the writer;

ensure IsTerminalStreamEvent recognizes the provider’s exact wire shape;

add an end-to-end regression test for Responses + passThroughBody where response.completed is forwarded without an injected STREAM_ERROR.

Recent unstable already recognizes response.completed both from StreamEvent.Type and JSON data.type via #2068. If you are already on a version containing that change, could you share the exact final SSE frame emitted by the provider and the AxonHub version/commit? That would help determine whether this is an unrecognized event shape or an event-delivery issue in the fan-out path.

So I think this should be handled as a separate pass-through regression rather than extending this PR’s Chat Completions synthesis logic.

Issue #1507, opened 2026-04-26
[Bug/错误]: UI caps model association priority at 10, but docs and backend allows up to 100


Sewer56:
The model association priority field in the AxonHub Admin UI is capped at 10, but the backend schema and documentation allow values up to 100. This creates an inconsistency where higher priority values can be set via API but not via the UI.

## Steps to Reproduce

1. Navigate to **Models** in Admin UI
2. Open a model's **Association** dialog
3. Try to set a rule's priority to any value above 10

**Expected**: UI and backend/schema use the same max priority value
**Actual**: UI caps at 10 while backend/schema cap at 100, creating an inconsistency

## Root Cause

In `frontend/src/features/models/components/models-association-dialog.tsx`, the priority field is capped at 10 in three locations:

1. **Zod form schema** - `.max(10)`
2. **HTML input attribute** - `max={10}`
3. **onChange handler** - `Math.max(0, Math.min(10, ...))`

Additionally, `handleAddAssociation()` auto-assigns priority to newly added rules using `Math.min(lastPriority + 1, 10)`, preventing the auto-increment from exceeding 10.

## Origin

PR #344 (Dec 24, 2025) - "fix: model association priority and ui align" - **introduced both `.max(10)` in the dialog AND `.max(100)` in `schema.ts` in the same PR**, creating the inconsistency from the start.

Prior to PR #344, priority had **no max constraint** (`z.number().optional().default(0)` from PR #330).

## Documentation vs UI Contradiction

The [Model Management Guide](https://github.com/looplj/axonhub/blob/unstable/docs/en/guides/model-management.md) explicitly recommends:

> **Priority Settings** - Primary channels: 0-10, backup channels: 10-100

The documentation clearly intends for priority values above 10 to be used for backup and emergency tiers.

### Files to modify

**`frontend/src/features/models/components/models-association-dialog.tsx`**

| Location              | Current                          | Change                      |
| --------------------- | -------------------------------- | --------------------------- |
| Form schema Zod       | `.max(10)`                       | `.max(100)`                 |
| HTML input `max` attr | `max={10}`                       | `max={100}`                 |
| onChange clamp        | `Math.min(10, ...)`              | `Math.min(100, ...)`        |
| handleAddAssociation  | `Math.min(lastPriority + 1, 10)` | `lastPriority + 1` (no cap) |

### Additional improvements

1. Extract max priority to a constant (e.g., `MAX_ASSOCIATION_PRIORITY = 100`) for maintainability
[...]

PR #2327, opened 2026-08-30
fix(models): align association priority cap with backend contract

### LINKED ISSUE #2323: [Bug]: Model association priority capped at 10 in UI, disabling Save with no feedback (#1507 not actually fixed)
## Description

The priority input in the model association dialog caps at `10`, while the backend contract and the official docs both allow `0-100`. This causes two cascading failures:

1. Typing `60` is **silently rewritten** to `10`, with no hint or error.
2. Any existing rule with `priority > 10` makes the whole form fail validation → **Save is permanently disabled with no visible reason**, and the value cannot be corrected through the UI (changing it back gets clamped to 10 again).

### On #1507 / #1581

Point 1 is what #1507 reported. #1581 (commit `aa95b16a`) claims to close it, but that commit **only extracted the literal `10` into a constant `MAX_ASSOCIATION_PRIORITY = 10` without changing the value to `100`**:

```diff
+const MAX_ASSOCIATION_PRIORITY = 10;
-        priority: z.number().min(0, '...').max(10, 'Priority cannot exceed 10'),
+        priority: z.number().min(0, '...').max(MAX_ASSOCIATION_PRIORITY, `Priority cannot exceed ${MAX_ASSOCIATION_PRIORITY}`),
```

So #1507 is **still unfixed**. Confirmed on current `unstable` HEAD (`4483c2e4`).

## Steps to Reproduce

**A. Priority cannot be set above 10**

1. Go to **Models** → any model → **Model Association**
2. Change a rule's priority to `60`
3. The input immediately snaps back to `10`

**B. Save permanently disabled with no feedback**

1. The model has an association rule with `priority > 10` (writable via GraphQL — see Notes)
2. Open the **Model Association** dialog
3. Save is disabled, and no error message appears anywhere in the UI
4. Try lowering that row's priority → the input clamps to `10`; any value above 10 is rejected

## Expected Behavior

- Priority accepts `0-100`, consistent with the backend contract and the docs
- Validation failures show an error message on the offending field instead of silently disabling Save

## Actual Behavior

- UI caps at `10`, making 91% of the documented range unreachable
- Validation failures are completely invisible

## Root Cause

`frontend/src/features/models/components/models-association-dialog.tsx`

| Location | Current | Note |
| --- | --- | --- |
| L32 | `const MAX_ASSOCIATION_PRIORITY = 10;` | should be `100` |
| L366 | `.max(MAX_ASSOCIATION_PRIORITY, ...)` | follows the constant |
| L1418 | `max={MAX_ASSOCIATION_PRIORITY}` | follows the constant |
| L1421 | `Math.max(0, Math.min(MAX_ASSOCIATION_PRIORITY, ...))` | follows the constant |
| L1411-1428 | priority `FormItem` has **no `<FormMessage />`** | every other field in this file (type/channelId/modelId/pattern) has one, so this field's validation errors are invisible |

Save button disabled condition (L1157):

```tsx
disabled={updateModel.isPending || updateModelSettings.isPending || isSettingsLoading || !form.formState.isValid}
```

`formState.isValid` covers the entire zod schema, so a single row with `priority > 10` disables Save permanently — and the missing `FormMessage` makes the cause invisible.

### Inconsistent with the backend contract

- `frontend/src/features/models/data/schema.ts:119` — `priority: z.number().min(0).max(100).optional().default(0)`
- `internal/objects/model.go:67` — `Priority int // Lower value = higher priority, default 0`
- `internal/server/gql/model.graphql:144` — `priority: Int` (no range constraint)
- `docs/en/guides/model-management.md` — "Primary channels: 0-10, Backup: 10-50, Emergency: 50-100"

## Verification

Ran the real zod schema from `unstable` HEAD (L32-432 of `models-association-dialog.tsx`, extracted verbatim) against real zod v4:

```
MAX_ASSOCIATION_PRIORITY = 10

typed "60"  -> stored 10   <-- SILENTLY CHANGED
typed "20"  -> stored 10   <-- SILENTLY CHANGED
typed "100" -> stored 10   <-- SILENTLY CHANGED

complete rule with priority=20 (backend-legal):
  associations.0.priority: Priority cannot exceed 10
  formState.isValid = false -> Save DISABLED
```

Present in every version, unfixed throughout:

| Version | Priority cap | priority field has FormMessage |
| --- | --- | --- |
| v0.9.43 | `.max(10, ...)` | no |
| v1.0.0-beta4 | `MAX_ASSOCIATION_PRIORITY = 10` | no |
| v1.0.0-beta7 | `MAX_ASSOCIATION_PRIORITY = 10` | no |
| unstable `4483c2e4` | `MAX_ASSOCIATION_PRIORITY = 10` | no |

## Suggested Fix

1. Set `MAX_ASSOCIATION_PRIORITY` to `100`, aligning with `schema.ts` and the docs
2. Add `<FormMessage />` to the priority `FormItem` so validation errors are visible
3. Update the cap number in the i18n error string
4. `handleAddAssociation` (L905-915) currently assigns the new rule `priority: lastPriority` — the **same** value as the current highest priority. Colliding priorities make match order fall through to `sourceRank` / insertion order (`internal/server/biz/model_settings_inheritance.go:319-323`). Suggest incrementing within the cap instead.

### Note on data provenance

None of the frontend write paths produce values above 10 (`modelAssociationToFormRow` / `formAssociationToModelAssociation` all pass through `assoc.priority ?? 0`; the batch-create dialog hardcodes `0`). `ModelAssociationInput.priority` is an unconstrained `Int` at the GraphQL layer with no range check in the resolver, which is how existing `> 10` values get written. After the fix such data becomes valid and editable with no migration needed — though it may be worth deciding whether the backend should enforce the same range.

## Environment

- Reproduced on: v0.9.43, v1.0.0-beta7, `unstable` HEAD `4483c2e4`
- Related: #1507 (same issue, incorrectly marked fixed by #1581)

ronronner02:
## Summary

The association dialog caps priority at `10`, but `modelAssociationSchema` allows `0-100` and the [Model Management Guide](https://github.com/looplj/axonhub/blob/unstable/docs/en/guides/model-management.md) recommends tiers up to 100 (primary 0-10, backup 10-50, emergency 50-100). That mismatch causes two failures:

1. Typing a value above 10 is **silently rewritten** to 10 — no error, no hint.
2. Any existing rule with `priority > 10` fails form validation, so **Save is permanently disabled with no visible reason**, and the value cannot be corrected in the UI because the input clamps back to 10.

The second one is the worse of the two: `disabled={... || !form.formState.isValid}` gates Save on the whole schema, and the priority `FormItem` is the only field in this file without a `<FormMessage />`, so the failure is invisible. A model configured through the API with tiered priorities becomes read-only in the UI.

Note: #1507 reported the cap and was closed by #1581, but that commit only extracted the literal `10` into `MAX_ASSOCIATION_PRIORITY = 10` without changing the value, so the issue is still present on `unstable`.

## Changes

`frontend/src/features/models/components/models-association-dialog.tsx` (4 lines)

- `MAX_ASSOCIATION_PRIORITY`: `10` → `100`. The zod schema, the `max` attribute and the `onChange` clamp all reference this constant.
- Added `<FormMessage />` to the priority field so validation errors surface.
- `handleAddAssociation` assigned `priority: lastPriority` — the same value as the current highest-priority rule. Colliding priorities make match order fall through to `sourceRank` and insertion order (`internal/server/biz/model_settings_inheritance.go:319-323`), which is not predictable from the UI. Now increments within the cap; the empty-list seed changes to `-1` so the first rule still gets `0`.

## Verification

Ran the real schema region of this file (L32-432, extracted verbatim) against real zod v4:

**Before**
```
MAX_ASSOCIATION_PRIORITY = 10
typed "60" -> stored 10          # silently changed
rule priority=20 (backend-legal): associations.0.priority: Priority cannot exceed 10
  formState.isValid = false -> Save DISABLED
```

**After**
```
MAX_ASSOCIATION_PRIORITY = 100
typed "60" -> 60 ; "100" -> 100 ; "101" -> 100 ; "-5" -> 0 ; "abc" -> 0
rule priority=20: no errors -> Save enabled
tiers 0 / 10 / 50 / 100 all accepted; 101 and -1 still rejected
new-rule priority: [] -> 0 ; [0] -> 1 ; [20] -> 21 ; [99] -> 100 ; [100] -> 100 (capped)
```

An incomplete new row still disables Save (it genuinely lacks channel/model) — the difference is the errors are now visible.

Toolchain matching `build.yml` (Node 22, pnpm 10):

```
pnpm install --frozen-lockfile   Done in 15.2s
npx tsc --noEmit                 exit 0, no type errors
pnpm build                       built in 1m 12s, exit 0
```

`pnpm lint` reports 252 problems (159 errors, 93 warnings) — **identical on unpatched `unstable`**, so this change adds none.

Not run: repo E2E (needs a Go backend). No formatting commit: `prettier` rewrites 169 lines of this file on unpatched `unstable` too, so reformatting would bury a 4-line change.

## Notes

`ModelAssociationInput.priority` is an unconstrained `Int` in GraphQL (`internal/server/gql/model.graphql:144`) with no range check in the resolver, which is how `> 10` values get in. This PR is UI-only; existing data becomes valid and editable with no migration. Whether the backend should enforce the same `0-100` range seems worth deciding separately.

Closes #2323

## Summary by CodeRabbit

* **New Features**
  * Association priorities can now be set up to 100.
  * New associations automatically receive the next available priority, capped at 100.
  * Priority validation messages are displayed directly for each field.

### DIFF
--- a/frontend/src/features/models/components/models-association-dialog.tsx
+++ b/frontend/src/features/models/components/models-association-dialog.tsx
@@ -29,7 +29,7 @@ import { ModelAssociation, normalizeModelRoutingPolicyValue } from '../data/sche
 import { toast } from 'sonner';
 import { ChannelModelsList } from './channel-models-list';
 
-const MAX_ASSOCIATION_PRIORITY = 10;
+const MAX_ASSOCIATION_PRIORITY = 100;
 
 const requestFormatConditionOptions = [
   'openai/chat_completions',
@@ -908,11 +908,11 @@ export function ModelsAssociationDialog() {
     // Get the priority of the last rule (highest priority)
     const currentAssociations = form.getValues('associations') || [];
     const lastPriority =
-      currentAssociations.length > 0 ? Math.max(...currentAssociations.map((a) => a.priority ?? 0)) : 0;
+      currentAssociations.length > 0 ? Math.max(...currentAssociations.map((a) => a.priority ?? 0)) : -1;
 
     append({
       type: 'channel_model',
-      priority: lastPriority,
+      priority: Math.min(lastPriority + 1, MAX_ASSOCIATION_PRIORITY),
       disabled: false,
       whenEnabled: false,
       whenCondition: DEFAULT_WHEN_CONDITION,
@@ -1423,6 +1423,7 @@ function AssociationRow({ index, form, isDeveloperMode, channelOptions, allModel
                   placeholder='0'
                 />
               </FormControl>
+              <FormMessage />
             </FormItem>
           )}
         />


llc1123 (commented):
## Review summary

我确认这个 PR 解决的问题是真实存在的：文档和共享前端 schema 使用 `0..100`，但关联模型对话框此前仍把 priority 限制为 `0..10`。不过当前实现还不能合并，主要原因如下。

### Required changes

1. **不要在校验前 clamp 输入值**

当前 `onChange` 仍执行：

```ts
Math.max(0, Math.min(MAX_ASSOCIATION_PRIORITY, Number(e.target.value) || 0))
```

因此 `101` 会被静默改成 `100`，`-1` 会被静默改成 `0`，Zod 的错误信息永远不会显示。请保留用户输入，让 schema/FormMessage 负责报告越界值，并补充 `0`、`100`、`101`、`-1` 的边界测试。由于 GraphQL 使用 `Int`，也请明确是否应拒绝小数。

2. **明确达到 100 后的重复 priority 策略**

当前 append 逻辑在已有最大值为 `100` 时仍返回 `100`，会产生重复 priority。现有 load balancer 语义允许相同 priority 组成同一组，因此这未必是 bug，但 PR 需要明确产品意图：

- 如果 priority 必须唯一，请选择下一个未使用值或在上限处阻止 append；
- 如果相同 priority 是有意的，请在文档/测试中明确“上限饱和后进入同一 priority tier”。

### Contract note

当前后端 `objects.ModelAssociation.Priority`、GraphQL `Int`、model settings 校验和 backup restore 路径都没有 `0..100` 的服务端范围校验。本 PR 至少应明确这是前端显示/输入契约；如果 `0..100` 是真正的后端契约，还需要同步覆盖 model、developer settings 和 restore 写入路径。

结论：问题真实，但上述两项实现/策略需要先补齐，再进入合并门禁。

PR #2332, opened 2026-08-30
Add MiniMax image generation support

octo-patch:
Reason: Add MiniMax image generation support to the image endpoint.

Added MiniMax image-generation request and response handling for global and China endpoints, including image models, prompt options, dimensions, subject references, URL or base64 output, and bearer authentication. Registered image generation as a MiniMax channel endpoint while preserving OpenAI-compatible chat behavior.

Checks: `go test ./llm/transformer/minimax ./llm/transformer/openai ./internal/server/biz` (dependency download in progress)

## Summary by CodeRabbit

* **New Features**
  * Added MiniMax support for OpenAI-compatible chat completions.
  * Added MiniMax image generation with image URL and Base64 responses.
  * Added image-generation options for subject references, aspect ratios, custom dimensions, seeds, and prompt optimization.
  * MiniMax endpoints are selected automatically for chat and image-generation requests.

### DIFF
--- a/internal/server/biz/channel_endpoint.go
+++ b/internal/server/biz/channel_endpoint.go
@@ -102,6 +102,11 @@ var openAIChatOnlyDefaultEndpoints = []objects.ChannelEndpoint{
 	{APIFormat: llm.APIFormatOpenAIChatCompletion.String()},
 }
 
+var minimaxDefaultEndpoints = []objects.ChannelEndpoint{
+	{APIFormat: llm.APIFormatOpenAIChatCompletion.String()},
+	{APIFormat: llm.APIFormatOpenAIImageGeneration.String()},
+}
+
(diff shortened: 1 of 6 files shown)


octo-patch:
I fixed the image-generation seed mapping that caused the Go compile failure and pushed a new commit. I ran `go test ./transformer/openai ./transformer/minimax` from the `llm` module.

looplj:
Please help to confirm if the AI review comments are reasonable.

octo-patch:
Three comments are valid: the default model should map to image-01, incoming b64_json should be translated to MiniMax base64, and a custom versioned endpoint path should not produce a duplicate /v1. The seed issue is already fixed in 610f464; enforcing HTTPS only here would conflict with repository-wide configurable base URLs, and the supplied MiniMax response contract does not require preserving an id.

PR #2342, opened 2026-09-01
fix(anthropic): preserve client cache_control breakpoints to keep prompt cache hits

Azhc:
## Problem

The auto-optimization pipeline (`optimizeCacheControl`) clears all client-provided `cache_control` breakpoints and re-plans them. The second message anchor is placed at a fixed window boundary (~20 blocks from the end, `pickWindowAnchorIndex`), so **as a conversation grows, the anchor lands on a different block in every request**.

Anthropic prompt caching only hits when the prefix up to a breakpoint is byte-identical to a previously cached entry. A moving breakpoint invalidates the prefix hash on every turn, so every request pays a **full cache write** and `cache_read_input_tokens` stays 0 — even though the actual conversation prefix is unchanged.

## Evidence (real deployment, single user, single conversation)

From `request_executions` on one channel over a few hours: **139/148 requests had `cache_read_input_tokens = 0`**, all paying full cache writes (~24K tokens each).

Controlled replay directly against the upstream (identical headers, only the body varying):

| Scenario | cache_read |
|---|---|
| Same request body sent twice | **24232** (full hit) |
| Same conversation +2 messages, breakpoints re-planned (old behavior) | **0** (full rewrite) |
| Same conversation +2 messages, breakpoint kept at original position | **24419** (hit) |

The upstream relay/cache itself works fine — the moving breakpoint is the sole cause.

Secondary issue: client-set `TTL` values (`1h`/`5m`) were silently wiped by the strict re-plan; they are now preserved.

## Fix

- **Client message breakpoints are preserved verbatim** (position + TTL) — they are part of the cache prefix key and must be byte-stable across turns
- Structural anchors (tools/system last) are only filled when the section has **no** breakpoint at all, so client anchor choices are respected and the total is not pushed over the limit unnecessarily
- When the total exceeds the 4-breakpoint API limit, **the earliest message breakpoints are trimmed** (they cover the shortest prefix, so removing them does not invalidate the remaining cached entries) instead of re-planning everything
- Requests with **zero** message breakpoints keep the existing auto-inject behavior unchanged

## Tests

- Updated `ensure_cache_control_test.go`, `cache_control_test.go`, `integration_test.go` to the preserve/trim semantics (kept all zero-breakpoint auto-inject tests as-is)
- `go test ./...` passes in `llm/`

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved client-defined cache-control settings for system prompts, tools, images, and messages.
  * Improved handling when cache-control breakpoints exceed the allowed limit by trimming only the earliest message breakpoints.
  * Retained valid client breakpoints instead of rebuilding or removing them unnecessarily.
  * Maintained structural cache anchors unless client-defined breakpoints are already present.
  * Improved cache-control handling for nested and raw message content.

### DIFF
--- a/llm/transformer/anthropic/cache_control_test.go
+++ b/llm/transformer/anthropic/cache_control_test.go
@@ -397,8 +397,10 @@ func TestOutboundTransformer_CacheControl(t *testing.T) {
 		require.NotNil(t, anthropicReq.System)
 		require.Len(t, anthropicReq.System.MultiplePrompts, 2)
 
-		// strict mode 下仅保留 system 最后一个结构锚点
-		require.Nil(t, anthropicReq.System.MultiplePrompts[0].CacheControl)
+		// 客户端声明的断点原样保留（含 TTL），不再清空重排
+		require.NotNil(t, anthropicReq.System.MultiplePrompts[0].CacheControl)
+		require.Equal(t, "ephemeral", anthropicReq.System.MultiplePrompts[0].CacheControl.Type)
+		require.Equal(t, "5m", anthropicReq.System.MultiplePrompts[0].CacheControl.TTL)
 
 		// Check second system prompt has cache control
 		require.NotNil(t, anthropicReq.System.MultiplePrompts[1].CacheControl)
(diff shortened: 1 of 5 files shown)


looplj:
帮忙确认下是否有实际验证过缓存效果

PR #2357, opened 2026-09-03
fix: preserve transformed Responses stream events

llc1123:
## Summary

- cache the event consumed by `responsesSessionStream.Next`
- return that cached event from `Current` instead of advancing the inner stream again
- add a regression test proving every wrapped event reaches the downstream consumer

## Root cause

The Responses session wrapper added in #2255 reads each inner stream event in `Next` so it can record completed sessions. Its `Current` method then called `inner.Current()` a second time.

That is destructive for transformed Responses streams: queue-backed implementations such as `responsesInboundStream` advance their cursor on every `Current()` call. The wrapper consumed sequence 0 internally, returned sequence 1 to the client, consumed sequence 2 internally, returned sequence 3, and eventually returned `nil` after consuming a single queued delta. The client never received `response.completed`, so the request ended as an incomplete stream.

Native Responses pass-through masked the bug because `passThroughChannelStream.Current()` returns a cached pointer and is idempotent.

## Impact

Affected when all of the following are true:

1. downstream API format is `openai/responses`
2. `stream=true`
3. response pass-through is not applied
4. the final Responses event stream has consumptive `Current()` semantics

This includes Responses requests transformed through Chat Completions, Anthropic, Gemini, or a non-pass-through Responses path. Both SSE and downstream Responses WebSocket use the same `Next`/`Current` contract.

Not affected:

- non-streaming Responses requests
- non-Responses downstream APIs
- native Responses streams with response pass-through applied

The incident appeared model-specific because the sampled GPT-5.6 traffic used native Responses pass-through and completed, while Kimi was transformed through `openai/chat_completions` with `passThroughApplied=false`. The model itself is not the boundary.

## Verification

The regression test failed before the fix with three expected events becoming only the second event plus `nil`. It passes after caching the current event.

```text
go test -race -shuffle=on -count=1 ./internal/server/orchestrator
ok github.com/looplj/axonhub/internal/server/orchestrator
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed streaming event handling so consumers receive each event exactly once and in the correct order.
  * Ensured the current event is cleared when the stream is exhausted.

* **Tests**
  * Added coverage verifying that all events in a response stream are preserved and forwarded unchanged.

### DIFF
--- a/internal/server/orchestrator/responses_session.go
+++ b/internal/server/orchestrator/responses_session.go
@@ -281,6 +281,7 @@ type responsesSessionStream struct {
 	store       *responsesSessionStore
 	requestBody []byte
 	inner       streams.Stream[*httpclient.StreamEvent]
+	current     *httpclient.StreamEvent
 	chunks      []*httpclient.StreamEvent
 	chunksBytes int
 	terminal    bool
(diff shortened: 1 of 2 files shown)


cjjdaq:
image

looplj:
https://github.com/looplj/axonhub/pull/2365
是不是和这个重复了

cjjdaq:
> #2365 是不是和这个重复了

对，是那个问题。

llc1123:
确认重复。#2365 的实现和回归测试与本 PR 等价，并且已经合并到 `unstable`。为避免保留重复 PR，这里关闭 #2357。感谢确认。

llc1123:
补充说明一下：#2357 比 #2365 早提交约 5 小时 40 分钟，#2365 的描述里也明确把 #2357 列在 Related 中；实际对比后，两者的修复逻辑和回归测试基本等价。

在没有说明技术取舍理由的情况下，最终合并后提交的重复 PR、关闭先提交且已经完成定位和验证的 PR，这种处理对先发现问题并提交修复的贡献者并不公平，也会影响后续参与贡献的积极性。

希望以后遇到重复 PR 时：

1. 优先评估和合并先提交、且已经有完整验证的 PR；
2. 如果确实选择后提交的版本，请公开说明具体技术差异和选择理由；
3. 对原始问题定位、复现证据和修复贡献保留明确的来源说明与 credit。

这次我会保持 #2357 关闭，但希望维护流程后续能更透明、公平一些。

looplj:
好的，理解。这个因为是看到了 issue ，然后看到原作者提交的 fix pr 所以没看其他的 pr，抱歉。

PR #2372, opened 2026-09-03
fix(openai): prevent zero-status errors from returning HTTP 200

### LINKED ISSUE #2083: [Bug/错误]:  Responses WebSocket streaming errors are returned as HTTP 200 and silently swallowed by OpenAI-compatible clients
### Pre-submission Checklist / 提交前检查

- [x] I have searched the [existing issues](https://github.com/looplj/axonhub/issues) and this bug has not been reported / 我已搜索现有 issues，此 Bug 尚未被报告
- [x] I have read the [documentation](https://github.com/looplj/axonhub#readme) / 我已阅读文档
- [x] I am using the latest version of AxonHub / 我正在使用 AxonHub 的最新版本

### FAQ Check / FAQ 检查

- [x] I have checked the [FAQ](https://github.com/looplj/axonhub/issues/1374) and this bug is not covered there / 我已查阅 FAQ，此 Bug 未被覆盖

### Bug Description / Bug 描述

When a Codex channel uses OpenAI Responses WebSocket transport, a top-level upstream error received before the first streaming response event can be returned to an OpenAI Chat Completions client as `HTTP 200 OK` with an `application/json` error body.

For example, an upstream `context_length_exceeded` error is returned over the wire as:

```http
HTTP/1.1 200 OK
Content-Type: application/json; charset=utf-8

{"error":{"code":"context_length_exceeded","message":"Your input exceeds the context window of this model. Please adjust your input and try again.","type":"","param":"input"}}
```

This request was made with `stream: true`. OpenAI-compatible streaming clients treat every 2xx response as a successful SSE response. Because the body is raw JSON rather than SSE, clients such as `@ai-sdk/openai-compatible` receive no data frames and no terminal event. The upstream error is therefore silently swallowed.

In OpenCode, the affected generation advances/exits without displaying a provider error. It also cannot classify the failure as a context overflow, so automatic overflow compaction/recovery is not triggered.

Environment:

- AxonHub: `v1.0.0-beta5`
- Installed commit: `d061ac7df6aef0c5ec6cdfa9dc5002546a1c5a57`
- Also present on current `unstable` at the time of reporting: `7d095b6364f4e765d687d22f2f1a7c6536de92ad`
- Inbound endpoint: `POST /v1/chat/completions`
- Inbound format: OpenAI Chat Completions with `stream: true`
- Outbound channel: Codex using OpenAI Responses WebSocket transport
- Client: OpenCode with `@ai-sdk/openai-compatible`

Probable cause:

1. `llm/transformer/openai/responses/outbound_stream.go` handles `StreamEventTypeError` by returning an `llm.ResponseError` without a `StatusCode`, leaving it as zero.
2. `llm/transformer/openai/inbound.go` copies that zero value to `httpclient.Error.StatusCode`.
3. `internal/server/api/chat.go` calls `c.JSON(httpErr.StatusCode, ...)`, effectively `c.JSON(0, ...)`.
4. Gin ignores non-positive status codes and retains its default `200` status when writing the JSON body.
5. The OpenAI-compatible AI SDK selects its successful SSE response handler because the status is 2xx, then receives raw JSON instead of SSE frames.

This is separate from an error that occurs after SSE headers have already been committed. In the reproduced case, AxonHub logged `Error processing chat completion`, meaning the failure was available before the client-facing stream was opened and a non-2xx JSON response could still have been sent.

### Steps to Reproduce / 复现步骤

1. Configure a Codex channel with `OpenAI (Responses WebSocket)` transport.
2. Send an OpenAI Chat Completions streaming request whose synthetic input exceeds the selected model context window.
3. Capture the client-facing HTTP status, content type, and body.

Example using only synthetic repeated text:

```bash
export AXONHUB_URL='https://your-axonhub.example.com'
export AXONHUB_API_KEY='redacted'

bun -e '
const content = "token ".repeat(450000)
const response = await fetch(`${process.env.AXONHUB_URL}/v1/chat/completions`, {
method: "POST",
headers: {
Authorization: `Bearer ${process.env.AXONHUB_API_KEY}`,
"Content-Type": "application/json",
Accept: "text/event-stream",
},
body: JSON.stringify({
model: "gpt-5.6-sol",
messages: [{ role: "user", content }],
max_tokens: 32,
stream: true,
reasoning_effort: "low",
}),
})

console.log(response.status, response.statusText)
console.log(response.headers.get("content-type"))
console.log(await response.text())
'
```

Observed result:

```text
200 OK
application/json; charset=utf-8
{"error":{"code":"context_length_exceeded","message":"Your input exceeds the context window of this model. Please adjust your input and try again.","type":"","param":"input"}}
```

For comparison, the same synthetic request with `stream: false` returns a non-2xx response (`500` in the tested version), so the silent-client behavior is specific to the streaming path.

With OpenCode, the corresponding local log contains the stream start followed by the next loop step and loop exit, but no provider error:

```text
message=stream providerID=axonhub modelID=gpt-5.6-sol
message="llm runtime selected" llm.runtime=ai-sdk llm.provider=axonhub
message=loop step=1
message="exiting loop"
```

### Expected Behavior / 期望行为

If the upstream error is available before client-facing SSE headers are committed, AxonHub should return a non-2xx OpenAI-compatible JSON error response.

For `context_length_exceeded`, the expected response is preferably:

```http
HTTP/1.1 400 Bad Request
Content-Type: application/json

{
"error": {
"message": "Your input exceeds the context window of this model. Please adjust your input and try again.",
"type": "invalid_request_error",
"code": "context_length_exceeded",
"param": "input"
}
}
```

This allows standard OpenAI-compatible clients to raise an API error and preserve `error.code`. OpenCode can then classify the failure as context overflow and run its automatic compaction/recovery path.

AxonHub should never pass a zero status code to `c.JSON`. A defensive fallback should map any `llm.ResponseError` with `StatusCode == 0` to a valid non-2xx status, even whenthe provider-specific error has no explicit HTTP status.

Suggested regression coverage:

- A top-level Responses WebSocket `error` event with `code: context_length_exceeded` before the first content event.
- OpenAI Chat Completions ingress with `stream: true`.
- Assert HTTP status is non-2xx, ideally `400`.
- Assert `Content-Type` is JSON.
- Assert `error.code`, `error.message`, and `error.param` are preserved.
- Assert the response is not emitted as `HTTP 200` raw JSON or an empty SSE stream.

### Error Message / 错误消息

```shell
Sensitive identifiers have been replaced with placeholders.


{"level":"WARN","time":"2026-07-26T11:20:47Z","logger":"axonhub","message":"request process failed","error":"error: Your input exceeds the context window of this model. Please adjust your input and try again., code: context_length_exceeded","channel_id":"<redacted>","channel_name":"<redacted>","model_id":"gpt-5.6-sol","trace_id":"<redacted>","request_id":"<redacted>","operation_name":"POST /v1/chat/completions"}
{"level":"ERROR","time":"2026-07-26T11:20:48Z","logger":"axonhub","message":"Error processing chat completion","error":"failed to stream request: error: Your input exceeds the context window of this model. Please adjust your input and try again., code: context_length_exceeded","trace_id":"<redacted>","request_id":"<redacted>","operation_name":"POST /v1/chat/completions"}


Sanitized client-facing wire capture:


{
"status": 200,
"statusText": "OK",
"contentType": "application/json; charset=utf-8",
"bodyBytes": 175,
"error": {
"code": "context_length_exceeded",
"message": "Your input exceeds the context window of this model. Please adjust your input and try again.",
"type": "",
"param": "input"
}
}
```

### Operating System / 操作系统

Linux (Ubuntu/Debian) / Linux（Ubuntu/Debian）

### AxonHub Version / AxonHub 版本

v1.0.0-beta5

### Usage Scenario / 使用场景

API

### API Format / API 格式

OpenAI - Chat Completions (/v1/chat/completions) / OpenAI 聊天补全

### Channel Type / 渠道类型

OpenAI / OpenAI

### Logs / 日志

```shell

```

### Additional Context / 其他补充信息

_No response_

Solaris-star:
## Summary

- normalize OpenAI inbound `httpclient.Error` values whose status is missing or outside the HTTP error range
- map status-less `llm.ResponseError` values to `502 Bad Gateway` while preserving the provider error body
- add regression coverage for both error paths

A provider can report a top-level stream error without an HTTP status. Passing status `0` to Gin leaves its default `200 OK`, so OpenAI-compatible streaming clients select the SSE parser and then see an empty stream instead of the JSON provider error. The fallback is `502` because these errors originate from an upstream provider. Existing valid 4xx/5xx statuses remain unchanged.

Fixes #2083

## Tests

- `cd llm && go test ./transformer/openai -run ^'TestInboundTransformer_TransformError$' -count=1`\n- `cd llm && go test ./transformer/openai -count=1`\n- `cd llm && go test ./... -count=1`\n- `go test ./internal/server/api -count=1`

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized invalid or missing HTTP error status codes to **502 Bad Gateway**.
  * Preserved original error response bodies while ensuring errors use valid HTTP status ranges.
  * Correctly handles nil and typed-nil errors, returning a **500 Internal Server Error** response.
  * Maps `context_length_exceeded` errors without a status code to **400 Bad Request**.
  * Preserves valid 4xx and 5xx error statuses.

### DIFF
--- a/llm/transformer/openai/inbound.go
+++ b/llm/transformer/openai/inbound.go
@@ -194,6 +194,15 @@ func (t *InboundTransformer) TransformError(ctx context.Context, rawErr error) *
 		}
 	}
 
+	llmErr, isLLMError := errors.AsType[*llm.ResponseError](rawErr)
+	if isLLMError && llmErr == nil {
+		return &httpclient.Error{
+			StatusCode: http.StatusInternalServerError,
+			Status:     http.StatusText(http.StatusInternalServerError),
+			Body:       xjson.MustMarshal(&OpenAIError{Detail: llm.ErrorDetail{Message: "An unexpected error occurred", Type: "unexpected_error"}}),
+		}
+	}
+
 	if errors.Is(rawErr, transformer.ErrInvalidModel) {
 		return &httpclient.Error{
 			StatusCode: http.StatusUnprocessableEntity,
@@ -223,7 +232,7 @@ func (t *InboundTransformer) TransformError(ctx context.Context, rawErr error) *
 		}
 	}
 
-	if llmErr, ok := errors.AsType[*llm.ResponseError](rawErr); ok {
+	if isLLMError {
 		statusCode := llmErr.StatusCode
 		if statusCode == 0 && llmErr.Detail.Code == "context_length_exceeded" {
 			statusCode = http.StatusBadRequest
--- a/llm/transformer/openai/inbound.go
+++ b/llm/transformer/openai/inbound.go
@@ -202,7 +202,7 @@ func (t *InboundTransformer) TransformError(ctx context.Context, rawErr error) *
 	}
 
 	if httpErr, ok := errors.AsType[*httpclient.Error](rawErr); ok {
-		return httpErr
+		return normalizeHTTPErrorStatus(httpErr)
 	}
 
 	// Handle validation errors
@@ -215,9 +215,14 @@ func (t *InboundTransformer) TransformError(ctx context.Context, rawErr error) *
 	}
 
 	if llmErr, ok := errors.AsType[*llm.ResponseError](rawErr); ok {
+		statusCode := llmErr.StatusCode
+		if statusCode < http.StatusBadRequest || statusCode > 599 {
+			statusCode = http.StatusBadGateway
+		}
+
 		return &httpclient.Error{
-			StatusCode: llmErr.StatusCode,
-			Status:     http.StatusText(llmErr.StatusCode),
+			StatusCode: statusCode,
+			Status:     http.StatusText(statusCode),
 			Body:       xjson.MustMarshal(&OpenAIError{Detail: llmErr.Detail}),
 		}
 	}
@@ -227,4 +232,16 @@ func (t *InboundTransformer) TransformError(ctx context.Context, rawErr error) *
 		Status:     http.StatusText(http.StatusInternalServerError),
 		Body:       xjson.MustMarshal(&OpenAIError{Detail: llm.ErrorDetail{Message: rawErr.Error(), Type: "internal_server_error"}}),
 	}
+}
+
+func normalizeHTTPErrorStatus(httpErr *httpclient.Error) *httpclient.Error {
+	if httpErr.StatusCode >= http.StatusBadRequest && httpErr.StatusCode <= 599 {
+		return httpErr
+	}
+
+	normalized := *httpErr
+	normalized.StatusCode = http.StatusBadGateway
+	normalized.Status = http.StatusText(http.StatusBadGateway)
+
+	return &normalized
 }
--- a/llm/transformer/openai/inbound.go
+++ b/llm/transformer/openai/inbound.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"reflect"
 	"strings"
 
 	"github.com/looplj/axonhub/llm"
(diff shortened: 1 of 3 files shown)


Solaris-star:
Production-path validation after deploying this patch to a sanitized OpenAI-compatible gateway:

- 10 streaming requests against an intermittently failing upstream
- 5 successful SSE responses, all with `finish_reason` and `[DONE]`
- 5 upstream failures returned as HTTP 502
- 0 cases of HTTP 200 with a non-SSE/JSON error body
- direct-origin capture preserved the provider JSON error body under HTTP 502

The upstream itself is still unstable; this patch fixes the protocol-level symptom that made compatible clients report an empty stream instead of an upstream error.

PR #2395, opened 2026-09-04
fix(api): suppress stream errors after terminal events, close #2318

### LINKED ISSUE #2318: [Bug/错误]: Responses 流已下发 response.completed 后上游断开，SSE writer 追加的 error 帧被 OpenAI SDK 直接 throw，已完整送达的响应被客户端作废
### Bug Description / Bug 描述

**Scenario**: `/v1/responses`, `stream=true`, OpenAI Responses channel with `passThroughBody: true`, OpenAI-SDK-based client (pi, Codex, etc.). The upstream delivers the full event sequence including `response.completed`, and then the TCP connection breaks later (chunked body truncated, `io.ErrUnexpectedEOF` on the gateway side). In the incident below the gap was 32 seconds — the client had already consumed `response.completed` and was busy running tool calls.

**Problem**: when `stream.Err() != nil` fires on a stream whose terminal event (`response.completed`) has **already been delivered**, `writeSSEStreamEnd` (internal/server/api/chat.go, unstable `4483c2e4`) unconditionally appends a bare `error` SSE frame:

```
event: error
data: {"error":{"message":"unexpected EOF","type":"server_error","code":""},"request_id":""}
```

The openai SDK (v6.x, `core/streaming.js:49`) **throws an APIError on any data frame containing `data.error`, without yielding it to the consumer**. The client never sees the frame as an event — the error surfaces as a thrown `APIError("unexpected EOF")` mid-iteration, and the consumer discards the response it had already fully received.

Consequences (real incident, 2026-08-29):

- Gateway sent `response.completed` with full output + usage (33811 tokens) at 07:43:53.
- Upstream broke the connection at 07:44:25; gateway logged `Error in stream: unexpected EOF` and appended the `event: error` frame.
- Client (pi agent, openai SDK 6.40) threw `APIError: unexpected EOF`; the assistant message was marked `stopReason=error` and the **entire completed response was discarded** — the user saw the turn fail after 52 s of successful streaming.
- Aggregation in the DB even records `status: failed, error: {server_error, stream_error, unexpected EOF}` for a response the client fully received as `completed`.

Why the bare error frame is wrong *after* a terminal event: `response.completed` is self-contained (full response object, output, usage). Once delivered, the client already knows how the response ended. A trailing transport error cannot add information the client can act on — but the SDK contract turns it into a fatal, **non-retryable-looking** error that invalidates the completed response. This is the inverse quadrant of #2185 (terminal missing → synthesize incomplete_stream) and of #2030 (terminal delivered → don't mark cancelled): here the terminal **was** delivered, so the writer should stay silent and log.

The code already tracks `terminalSeen` in the writer loop (added in #2185) — it is simply not consulted in the `streamErr != nil` branch.

Two smaller gaps found on the same path (also affects non-pass-through):

1. `IsTerminalStreamEvent` (internal/server/orchestrator/inbound.go) does not recognize `response.failed` / `response.incomplete` / `response.cancelled` as terminal, so `terminalSeen` stays false when the stream legitimately ends with `response.failed` (transform path), and the writer appends an `event: error` **after** a terminal failure event too.
2. `classifyStreamError` (llm/transformer/openai/responses/inbound_stream.go) has no `errors.Is(err, io.ErrUnexpectedEOF)` branch, so a mid-chunk truncation is classified `stream_error` with the raw Go text "unexpected EOF" instead of a stable `upstream_eof` code.

### Steps to Reproduce / 复现步骤

1. Create an OpenAI Responses channel with pass-through body enabled (`passThroughBody: true`).
2. Point it at an upstream (or a stub) that streams a normal Responses sequence: `response.created` → deltas → `response.completed`, then **holds the connection open** (client-side tool execution window) and finally closes it without the chunked terminator.
3. From an openai-SDK client, run a streaming `/v1/responses` request through AxonHub.
4. Observed: the client consumes `response.completed`, then the SDK throws `APIError: unexpected EOF` and the completed response is discarded. Gateway log shows `Error in stream: unexpected EOF`.

Writer-level repro (Go test, applies to unstable `4483c2e4`):

```go
// stream that yields response.completed, then returns a non-nil Err
func TestWriteSSEStream_ResponsesErrorAfterTerminalWithStreamErr(t *testing.T) {
	// ... mock stream: [response.created, response.completed], Err() = io.ErrUnexpectedEOF
	// passes today only because the trailing `event: error` frame is appended —
	// with an OpenAI SDK client attached, that frame becomes a thrown APIError
	// that discards the already-delivered completed response.
}
```

### Expected Behavior / 期望行为

In `writeSSEStreamEnd`, when `streamErr != nil` but a terminal event was already delivered to the client (`terminalSeen == true`), the writer should **suppress the trailing `error` frame** and log a warning for observability (the terminal event already told the client how the response ended):

```go
} else if terminalSeen {
	log.Warn(ctx, "Stream error after terminal event was delivered, suppressing trailing error event",
		log.Cause(streamErr))
} else {
	log.Error(ctx, "Error in stream", log.Cause(streamErr))
	c.SSEvent("error", formatErr(ctx, streamErr))
}
```

Plus the two smaller fixes:

- `IsTerminalStreamEvent`: recognize `response.failed` / `response.incomplete` / `response.cancelled` as terminal (both in the `Type` check and the JSON-`type` fallback case).
- `classifyStreamError`: map `io.ErrUnexpectedEOF` to `code: "upstream_eof"` (transport truncation, not a provider-level `stream_error`).

Note the distinction from #2316: that issue asks for a **better-shaped** error frame for Chat Completions streams that break **before** any terminal (missing code/request_id). This issue is about the opposite case — the terminal was **already delivered** and the appended frame actively breaks SDK clients. The fixes are complementary; both are needed.

### Error Message / 错误消息

```shell
# pi client side (openai SDK 6.40 APIError, message is all the client gets):
unexpected EOF

# gateway log:
Error in stream: unexpected EOF

# gateway DB (request aggregation, post-incident):
status: failed
error: {"type": "server_error", "code": "stream_error", "message": "unexpected EOF"}
usage: {"input_tokens": 33488, "output_tokens": 323, "total_tokens": 33811}
```

### Operating System / 操作系统

Docker / Docker

### AxonHub Version / AxonHub 版本

v1.0.0-beta7 (reproduced; code path verified unchanged on `unstable` `4483c2e4`)

### Usage Scenario / 使用场景

API

### API Format / API 格式

OpenAI - Responses (/v1/responses) / OpenAI Responses

### Channel Type / 渠道类型

OpenAI / OpenAI

### Logs / 日志

```shell
# gateway (docker logs), 32s after response.completed was delivered:
Error in stream: unexpected EOF

# client session record (pi agent):
stopReason: "error"
errorMessage: "unexpected EOF"
rawStopReason: "completed"   # response.completed WAS delivered and processed
usage: {input: 33488, output: 323, total: 33811}  # complete
```

### Additional Context / 其他补充信息

- The full post-incident forensics: the 0-event `terminal_error` absence in the client's provider event log proves the client never received the error frame as a stream event — it only surfaced through the SDK's `data.error` → throw path.
- Related: #2030 (terminal delivered + client cancel → don't mark cancelled; same quadrant analysis, "pass-through canceled" mentioned in the commit), #2185 (`terminalSeen` introduced), #2316 (pre-terminal truncation error shape for Chat Completions), #1421 / #2083 (error shapes breaking client retry logic).
- The writer already has everything needed: `terminalSeen` is computed in both `writeSSEStream` loops; only the `streamErr != nil` branch ignores it.

looplj:
## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented spurious stream errors when a terminal response is followed by an unexpected disconnect.
  * Preserved completed response data and terminal details across connection closures.
  * Consistently handled completed, failed, incomplete, canceled, and standalone error responses.
  * Prevented duplicate terminal events from being emitted.
  * Retained final status and error details even when response aggregation encounters an error.
  * Improved upstream connection error classification and incomplete-response reason mapping.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -23,6 +23,8 @@ run:
 linters:
   default: all
   disable:
+    # Protocol strings may require British spelling (e.g. cancelled).
+    - misspell
     - prealloc
     - exhaustive
     - containedctx
@@ -157,11 +159,6 @@ linters:
       line-length: 180
       # tab width in spaces. Default to 1.
       tab-width: 1
-    misspell:
-      # Correct spellings using locale preferences for US or UK.
-      # Default is to use a neutral variety of English.
-      # Setting locale to US will correct the British spelling of 'colour' to 'color'.
-      locale: US
     nakedret:
       # make an issue if func has more lines of code than this setting and it has naked returns; default is 30
       max-func-lines: 30
--- a/internal/server/api/chat.go
+++ b/internal/server/api/chat.go
@@ -362,6 +362,10 @@ func writeSSEStreamEnd(
 ) {
 	switch {
 	case terminalSeen:
+		if streamErr != nil {
+			log.Warn(ctx, "Stream error after terminal event was delivered, suppressing trailing error event",
+				log.Cause(streamErr))
+		}
 	case errors.Is(ctx.Err(), context.Canceled):
 		*clientDisconnected = true
 
--- a/internal/server/api/chat_test.go
+++ b/internal/server/api/chat_test.go
@@ -1193,6 +1193,44 @@ func TestWriteSSEStream_TerminalEventWinsDeadlineRace(t *testing.T) {
 	require.NotContains(t, w.Body.String(), "event:error")
 }
 
+func TestWriteSSEStream_ResponsesErrorAfterTerminalIsSuppressed(t *testing.T) {
+	tests := []struct {
+		name      string
+		keepAlive SSEKeepAliveConfig
+	}{
+		{name: "without heartbeat"},
+		{
+			name: "with heartbeat",
+			keepAlive: SSEKeepAliveConfig{
+				Enabled:  true,
+				Interval: time.Hour,
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			w := httptest.NewRecorder()
+			c, _ := gin.CreateTestContext(w)
+			c.Request = httptest.NewRequest(http.MethodGet, "/", nil)
+			stream := &errorAfterStream{
+				items: []*httpclient.StreamEvent{
+					{Type: "response.created", Data: []byte(`{"type":"response.created"}`)},
+					{Type: "response.completed", Data: []byte(`{"type":"response.completed","response":{"status":"completed"}}`)},
+				},
+				err: fmt.Errorf("read body: %w", io.ErrUnexpectedEOF),
+			}
+
+			writeSSEStream(c, stream, FormatStreamError, tt.keepAlive, sseHeartbeatOpenAI)
+
+			body := w.Body.String()
+			require.Contains(t, body, "response.completed")
+			require.NotContains(t, body, "event:error")
+			require.NotContains(t, body, io.ErrUnexpectedEOF.Error())
+		})
+	}
+}
+
 func TestWriteSSEStream_RefreshesWriteDeadline(t *testing.T) {
 	w := httptest.NewRecorder()
 	c, _ := gin.CreateTestContext(w)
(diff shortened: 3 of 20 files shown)


huanglinfei091-cmd:
Hi @looplj — I noticed the `lint` job failed at `golangci-lint` in [this GitHub Actions run](https://github.com/looplj/axonhub/actions/runs/33902487055) for this PR.

I’m testing [RunBack](https://github.com/huanglinfei091-cmd/runback), an early alpha CLI that reconstructs a failed public Actions run locally with `act`. For this run, the useful question is whether the same golangci-lint issue is matched locally, or whether reconstruction stops with an honest blocked result.

```bash
runback doctor
runback https://github.com/looplj/axonhub/actions/runs/33902487055
```

The Linux amd64 binary and checksum are in the [v0.1.0-alpha release](https://github.com/huanglinfei091-cmd/runback/releases/tag/v0.1.0-alpha). If you try it, the verdict (`SAME_FAILURE`, `DIFFERENT_FAILURE`, or `REPLAY_BLOCKED`), Stage/Cause, and any install friction would be useful. Please do not post tokens or secrets. I’m asking for candid alpha evidence rather than a star or promotion; a failed reproduction is useful too.

PR #2401, opened 2026-09-05
fix(requests): keep API key filter visible without results

MrLiuGangQiang:
## Summary

- Keep the API key filter control mounted when a search returns no matching options.
- Allow the empty state to be displayed inside the filter instead of removing the control.

## Validation

- Not run (build and lint were not requested).

## Summary by CodeRabbit

* **Bug Fixes**
  * The caller/API-key filter now remains visible for authorized users whenever the corresponding table column is available, including while options are loading or unavailable.


MrLiuGangQiang:
@coderabbitai review

PR #2406, opened 2026-09-05
fix(playground): honor aliases when a channel is specified

moonjoke001:
## Summary
- Playground / TestChannel pin a channel via `SpecifiedChannelSelector`, which called `GetDirectModelEntries()`.
- That helper only kept `SupportedModels` ids, so auto-trim aliases (`kimi-k3`) and mappings (`deepseek-v4-pro`) were rejected even though the public model list (`allModelEntries`) showed them.
- Seen on `nv-api6` (`hideOriginalModels` + `autoTrimedModelPrefixes`) as `model kimi-k3 not supported in channel nv-api6`.

## Change
- `GetDirectModelEntries` now starts from `GetModelEntries()` (aliases included) and re-adds hidden original provider ids so channel tests can still target the upstream model name.

## Test plan
- [x] `go test ./internal/server/biz/ -run TestChannel_GetDirectModelEntries_IncludesAliasesAndHiddenOriginals`
- [x] `go test ./internal/server/orchestrator/ -run TestSpecifiedChannelSelector_Select`

### DIFF
--- a/internal/server/biz/channel_llm.go
+++ b/internal/server/biz/channel_llm.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"net/url"
 	"slices"
 	"strconv"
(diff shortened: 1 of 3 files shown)


moonjoke001:
Tracks https://github.com/looplj/axonhub/issues/2408 (Playground specified-channel rejects auto-trim / mapping aliases).

looplj:
还是 draft ，已经验证过了么

moonjoke001:
Closing in favor of #2493, which carries this fix (plus the channel test tokens, channel update guard, and OpenCode muse routing fixes) cherry-picked onto the current unstable base.

PR #2409, opened 2026-09-05
fix(openai): do not invent prompt_cache_key for compatible gateways

moonjoke001:
## Summary
- Anthropic inbound (Claude Code) always stores the session/trace id via `shared.WithSessionID`.
- `RequestFromLLM` then copied that id into Chat Completions `prompt_cache_key` for **every** OpenAI-compatible upstream.
- NVIDIA NIM (`integrate.api.nvidia.com`) rejects unknown `prompt_cache_key` with HTTP 400, so Claude Code saw empty/400 output while playground (AI SDK, no session fallback) succeeded on the same `kimi-k3` channel.

## Change
- `RequestFromLLM` no longer invents a cache key.
- Session fallback runs only for official `api.openai.com` hosts (Chat Completions and Responses).
- Explicit inbound `prompt_cache_key` is still forwarded.

## Test plan
- [x] `cd llm && go test ./transformer/shared/ ./transformer/openai/ ./transformer/openai/responses/`
- NVIDIA / other strict gateways: session present, no `prompt_cache_key` on the wire
- Official OpenAI: session fallback still applied

Related production incident: Claude Code session `02351f25-5398-4576-80a5-0e29cb79904d` → NVIDIA `Unsupported parameter(s): prompt_cache_key`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prompt cache keys are now automatically applied only when supported by the target OpenAI host.
  * Strict compatibility endpoints no longer receive unsupported prompt cache keys.
  * Explicitly provided prompt cache keys remain unchanged.

* **Tests**
  * Added coverage for supported OpenAI endpoints and compatibility hosts.
  * Verified that session IDs do not create prompt cache keys for unsupported endpoints.

### DIFF
--- a/llm/transformer/openai/outbound.go
+++ b/llm/transformer/openai/outbound.go
@@ -194,6 +194,7 @@ func (t *OutboundTransformer) TransformRequest(ctx context.Context, llmReq *llm.
 
 	// Convert to OpenAI Request format (this strips helper fields)
 	oaiReq := RequestFromLLM(ctx, llmReq, reasoningField)
+	applyPromptCacheKeyFallback(ctx, oaiReq, t.config.BaseURL)
 	//nolint:exhaustive // Checked.
 	switch t.config.PlatformType {
 	case PlatformOpenAI:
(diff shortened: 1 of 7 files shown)


looplj:
绝大部分应该是支持 prompt cache key 的吧，可选方案
1. 新增一个 options 是否去掉 prompt cache key
2. 给  nvidia 新增一个 Transformer 去掉  prompt cache key ，也是给 openai Transformer 新增一个参数。

而不是硬编码

PR #2418, opened 2026-09-06
fix(frontend): make channels table responsive without clipping actions

moonjoke001:
## Summary
- Drop the channels table `min-w-max` wrapper and clip horizontal overflow so the page no longer paints a 1px scrollbar (`overflow-x-hidden` + `table-fixed`).
- Hide secondary columns (quota / proxy / tags / createdAt) below xl/2xl, wrap the toolbar instead of scrolling it, and keep the action column at 11rem so edit / test / menu buttons are not clipped.

This is independent of the specified-channel alias fix (#2406). It targets the leftover horizontal scrollbar after #2366 added the OAuth quota column, and the action-button clipping that appeared once the table was constrained.

## Test plan
- [ ] Open `/channels` at ~1280px and ~1728px: no page-level or table-level horizontal scrollbar.
- [ ] Confirm edit / default-test / row-menu buttons are fully visible in the actions column.
- [ ] Widen past `xl` / `2xl` and confirm tags, createdAt, quota, and proxy columns reappear.
- [ ] Toolbar filters wrap instead of scrolling horizontally on a narrow window.

## Summary by CodeRabbit

* **Improvements**
  * Improved channel table responsiveness across screen sizes.
  * Long channel names, labels, and supported-model badges now wrap or truncate cleanly.
  * Adjusted column widths and hides lower-priority columns on narrower screens.
  * Toolbar controls now wrap onto multiple lines instead of requiring horizontal scrolling.
  * Reduced table cell padding for a more efficient layout.

### DIFF
--- a/frontend/src/features/channels/components/channels-columns.tsx
+++ b/frontend/src/features/channels/components/channels-columns.tsx
@@ -889,7 +889,7 @@ export const createColumns = (t: ReturnType<typeof useTranslation>['t'], canWrit
               </div>
             ),
             meta: {
-              className: 'text-center',
+              className: 'w-10 text-center',
             },
             enableSorting: false,
             enableHiding: false,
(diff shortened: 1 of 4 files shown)


looplj:
帮忙确认下是否有实际验证过

moonjoke001:
> 帮忙确认下是否有实际验证过

我的环境下用这个pr是正常的了

PR #2439, opened 2026-09-08
fix(frontend): gate project-scoped queries until persisted project is validated

brynne8:
## Context

This restores the fix path lost to account switching and closes the review findings on #2437 (now merged). Two independent problems were conflated; this PR handles both precisely.

### 1. A stale persisted project must never reach a project-scoped query

`selectedProjectId` is persisted per browser (`axonhub_selected_project_id`). After an account switch or a cold reload it can still carry a project the current user is **not** a member of, so `X-Project-ID`-carrying queries (`GetApiKeys`, `Users`) were denied (`input: permission denied`) — the "two ugly toasts" on the API keys page.

The merged fix cleared the selection on every auth transition. That was redundant once validation exists and it **defeats the remember-last-project feature** for returning users. This PR instead:

- keeps the persisted selection untouched on sign-in/sign-out;
- keeps protected content **unmounted until `me` has resolved and the selection is validated** against the current user's project memberships (`AuthGuard` gate + `useMe` clears it when stale), so a stale selection can never be sent; a returning user keeps their project.

### 2. Auxiliary `read_users` lookups must not noise up pages users can legitimately view

The default project **Developer** role grants `read_api_keys`, `write_api_keys`, `read_requests`, `write_requests` — **not** `read_users` (see `internal/server/biz/project.go`). Such a user must still list/create their own API keys, and the page never depends on `read_users` for that (`buildApiKeysQuery` omits the `user` sub-selection). But two places fetched the users list unconditionally and toasted a permission error when denied:

- API keys toolbar — already gated via `canViewCreators` + `disableAutoFetch` (#2437);
- **analytics filter-bar** — this PR adds the same gating and hides the user dropdown without `read_users`.

Members who do hold project-level `read_users` get a working read path via `ProjectMemberReadUsersRule` (#2437) once the request carries the correct `X-Project-ID`.

## Changes

- `auth-guard.tsx`: require a resolved `meData` (success, not error) before rendering protected children — an errored `me` no longer passes an unvalidated selection through (CodeRabbit).
- `auth.ts`: key the `['me']` cache by access token so an account switch on the 401-expiry path (which doesn't clear the query cache) cannot reuse another account's cached memberships (CodeRabbit).
- `auth.ts`: drop `clearSelectedProjectId()` from sign-in, OIDC exchange, and sign-out; `useMe`'s membership validation + the `AuthGuard` gate keep the invariant, and returning users keep their last project.
- `analytics/filter-bar.tsx`: run the users query only when the user has `read_users`, and hide the user filter otherwise.

## Verification

- `tsc --noEmit` passes.
- Branch is rebased on current `unstable`; net diff vs `unstable` is 5 files.
- Behavior matrix after this PR:
  | user | scenario | result |
  |---|---|---|
  | Developer (no `read_users`) | opens API keys page | lists & manages own keys, no denied toast |
  | Member with project `read_users` | opens API keys / users page | reads project users (requires correct `X-Project-ID`, sent) |
  | Any user | switches account on same browser | stale project never sent; first load shows loading until validation |
  | Returning user | re-logs in | last-used project is restored |

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved project access validation to prevent unauthorized project data requests.
  - Individual user data requests now respect the selected project.
  - Improved data refresh behavior when switching projects.
  - Restricted analytics user filters and related data loading based on permissions.
  - Improved authentication data refresh when access tokens change.
  - Automatically removes saved project selections that are no longer valid.
- **Tests**
  - Expanded coverage for project membership validation and project-scoped user access.

### DIFF
--- a/frontend/src/components/auth-guard.tsx
+++ b/frontend/src/components/auth-guard.tsx
@@ -3,6 +3,7 @@ import { useRouter } from '@tanstack/react-router';
 import { isAuthError } from '@/gql/graphql';
 import { useAuthStore } from '@/stores/authStore';
 import { useSelectedProjectId } from '@/stores/projectStore';
+import { isProjectSelectionValid } from '@/lib/project-membership';
 import { Skeleton } from '@/components/ui/skeleton';
 import { useMe } from '@/features/auth/data/auth';
 
@@ -24,10 +25,7 @@ export function AuthGuard({ children }: AuthGuardProps) {
   // descendant queries never carry a X-Project-ID the user has no membership in.
   // `meData` must be present: an unvalidated selection is never a reason to
   // render protected content, even when the `me` query itself errored.
-  const projectReady =
-    !!meData &&
-    (!selectedProjectId ||
-      (meData.projects ?? []).some((p) => p.projectID === selectedProjectId));
+  const projectReady = isProjectSelectionValid(meData, selectedProjectId);
 
   useEffect(() => {
     // If no token, redirect to sign-in
--- a/frontend/src/components/auth-guard.tsx
+++ b/frontend/src/components/auth-guard.tsx
@@ -22,10 +22,12 @@ export function AuthGuard({ children }: AuthGuardProps) {
   // the persisted project has been validated against the current user's
   // projects (and cleared when stale), keep showing the loading state so
   // descendant queries never carry a X-Project-ID the user has no membership in.
+  // `meData` must be present: an unvalidated selection is never a reason to
+  // render protected content, even when the `me` query itself errored.
   const projectReady =
-    !meData ||
-    !selectedProjectId ||
-    (meData.projects ?? []).some((p) => p.projectID === selectedProjectId);
+    !!meData &&
+    (!selectedProjectId ||
+      (meData.projects ?? []).some((p) => p.projectID === selectedProjectId));
 
   useEffect(() => {
     // If no token, redirect to sign-in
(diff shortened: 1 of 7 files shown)


brynne8:
暂时先别merge，我再改改

brynne8:
Both findings from the review are addressed in commit `f6ca409c` (plus the fixes already merged):

1. **`projectReady` treated `undefined meData` as ready** — fixed. It now requires a resolved `meData` before protected children render, so an errored `me` no longer lets an unvalidated persisted selection through (`frontend/src/components/auth-guard.tsx`).

2. **`['me']` query not isolated per auth token** — fixed. The query key is now `['me', accessToken]`, so the 401-expiry path (which does not clear the React Query cache) can no longer serve another account's cached memberships after re-login (`frontend/src/features/auth/data/auth.ts`).

While verifying the stale-project flow I also confirmed the default project **Developer** role grants no `read_users` (`internal/server/biz/project.go`), so this PR additionally stops the analytics filter-bar from firing (and toast-denying) the users lookup for users who genuinely lack `read_users` — the API keys page never depended on it. See the updated PR description for the behavior matrix.

brynne8:
Both findings from the latest review are addressed in `90fd5669`:

**1. Verify the project restriction after `privacy.Allow`** — added `TestProjectMemberReadUsersRuleAppliesProjectFilter`, an in-memory SQLite test that runs the real `User` schema policy (via `client.User.Query().All`). For the direct-scope and role-grant principals only the queried project's users are returned; the system-scope principal keeps the unfiltered view. A stray/dropped `WhereHasProjectUsersWith(projectID)` filter now fails these assertions.

**2. `role on a different project` fixture** — the fixture now grants membership in project 100 (the queried project) while the role is scoped to project 200, so the test actually exercises the different-project role branch instead of short-circuiting on the missing-membership check.

brynne8:
改完了。如果没问题可以merge一下。

PR #2495, opened 2026-09-18
fix(cline): keep quota fresh when usages ledger is rate limited

freebattle:
## Problem

Cline quota status stops refreshing for busy Cline Pass accounts.

`fetchUsageItems` rebuilds a 30-day cost ledger from `GET /api/v1/users/{id}/usages` on every quota check by paginating with `limit=100` up to 100 pages. Cline returns at most 200 items per page and rate limits rapid pagination. Once an account's usage grows past the rate limit, the ledger request returns `HTTP 429`, and `CheckQuota` aborts:

```
failed to read Cline usages: HTTP 429 Too Many Requests
```

The service then persists `error_code: check_failed` and pushes `next_check_at` out via `quotaErrorBackoff` (5min x2 capped at 8x = 40min), so the UI keeps showing a stale quota status.

Measured against a live account: ~33 sequential `/usages?limit=100` requests → `429`; a single request after cooldown → `200`, and `/plan/usage-limits` (the authoritative percent source) always returns `200`.

## Fix

- Treat the `/usages` ledger as best-effort. On failure it is marked unavailable (`usage_fetch.ledger_unavailable` / `ledger_error_code`), the per-window `cost_usage_*`/`used_cost_units` fields are omitted, and the official `usage-limits` percentages, reset times and balance are still reported.
- Request the documented page cap (`limit=200`, verified against the API) to halve the request count.
- Classify the ledger error without persisting the raw error string, which can embed credentials/URLs/cursors.

## Tests

- `TestCline_CheckQuota_UsageLedgerRateLimitedKeepsOfficialWindows` — 429 on `/usages` still yields official windows with `ledger_unavailable=true` and no cost fields.
- Updated the previous transport-error test to assert ledger errors are now non-fatal while still omitting sensitive values.
- Updated the happy-path page-size assertion.

`go test ./internal/server/biz/provider_quota/` and `go test ./internal/server/biz/ -run 'Cline|ProviderQuota|Quota'` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Quota checks now remain available when supplementary usage-history data cannot be retrieved.
  * Official usage-limit data continues to determine quota windows when the usage ledger is unavailable or incomplete.
  * Usage-history requests now support up to 200 records per page.
  * Quota details indicate when ledger data is unavailable and may include the relevant error information.
  * Cost breakdowns are omitted when usage-history data is unavailable or incomplete.


freebattle:
Thanks for the review — this one looks like a false positive. / 感谢审查，这条应该是误报。

**EN:** `costUnavailable` only reflects the `/usages` cost-ledger fetch, but the inactive branch never reads the ledger. `clineOfficialResetStateInactive` comes from the separate usage-limits response (`resetsAt: null`/empty), which means there is no active window for that period — so the window cost is authoritatively zero regardless of ledger availability. That's why we set `costSource = official_no_active_window` (not `..._ledger`) and still emit the zero cost. The contradictory case (inactive but `percentUsed > 0`) is already downgraded to `unavailable` just above, so no cost is emitted there. Happy to add a test for inactive + ledger-unavailable if you'd prefer.

**中文：** `costUnavailable` 只表示 `/usages` 成本账本拉取失败/被截断，而 inactive 分支根本不读取账本。`clineOfficialResetStateInactive` 来自另一个接口（usage-limits，`resetsAt: null`/空），含义是该窗口没有活跃期，因此该窗口成本在定义上就是 0，与账本是否可用无关。这也是我们把 `costSource` 标为 `official_no_active_window`（而不是 `..._ledger`）并仍然输出 0 成本的原因。真正矛盾的情况（inactive 但 `percentUsed > 0`）在上方已经被降级为 `unavailable`，不会输出成本。如果你希望的话，我可以补一个 “inactive + 账本不可用” 的测试来锁定该行为。

PR #2521, opened 2026-09-22
fix(dashboard): resolve empty fastestModels via LEFT JOIN

brynne8:
Fixes the fastestModels query which always returned empty results.

## Problem
The query used INNER JOIN on the `models` table, which is only populated when users manually create model records in the Models page. In most deployments this table is empty, causing the fastestModels endpoint to always return `[]`.

## Solution
- Change to LEFT JOIN so rows without model directory entries are included
- Use COALESCE to fall back to raw model_id when m.name is NULL
- Update test assertions to match the new query structure

## Testing
- All existing tests pass with updated assertions
- The change maintains the same result schema (modelName is always non-null)
- Frontend requires no changes

## Summary by CodeRabbit

- **Bug Fixes**
  - Model throughput results now remain available when a model name is missing.
  - In these cases, the model ID is shown as a fallback, improving the consistency of throughput reporting.


brynne8:
这个我认为可以合并，算是大重构的附带PR。这个问题早就存在，所以即便大PR不过，这个PR依旧有意义。

